PR #182, opened 2026-07-28
fix(events): DNT consent shim + share cards on every generated event page

PipFoweraker:
## What changed

One file of real source: `scripts/sync/sync-events.py`. Everything under `public/events/` is its output, re-rendered.

### 1. DNT coverage (TECH_DEBT A3)

`public/assets/js/analytics.js` is the consent shim that sets `localStorage.plausible_ignore` from Do-Not-Track or an explicit opt-out. It was loaded by **4** of 2,245 HTML files. A visitor arriving on a deep-linked event page was therefore counted by Plausible before the flag could ever be set — the privacy page makes a DNT-honouring promise that almost no page delivered. That is a truthfulness failure, not a nice-to-have.

The shim now sits in the `<head>` of every generated event page, **parser-blocking and above the deferred tracker**. That ordering is the whole mechanism: a plain script runs during parse, the Plausible tag is `defer`, so the flag is written before the pageview fires. Keeping it non-deferred also means the ordering survives someone later dropping `defer` from the tracker tag.

The shim file itself is untouched. It still creates no `<script>` element — `node scripts/test-analytics-optout.js` passes 17/17, including the three no-injection assertions that exist because that bug shipped once.

### 2. OpenGraph coverage (TECH_DEBT B4)

Added to the same template, derived from event data already in scope:

`og:type` (article), `og:site_name`, `og:title`, `og:description`, `og:url`, `og:image`, `twitter:card`, `twitter:title`, `twitter:description`.

`og:image` points at the existing site-wide card `https://pdoom1.com/assets/og-card.jpg` — the same one `index`/`about`/`press` already use. No new or per-event image is referenced, so the image-downscaling work happening in parallel can replace that file without touching these pages. `twitter:site` stays omitted, matching the note in `index.html` (handle not finalised).

### 3. Two supporting fixes in the same template

- **Attribute escaping.** Meta values are now whitespace-collapsed and HTML-escaped. This was already broken before this PR: the event titled `"Why Should I Trust You?": ...` contains double quotes, 6 titles contain `&`, and many arXiv-derived descriptions are multi-line, so `<meta name="description">` was emitting attribute values that spanned six physical lines and, for the quoted title, would have terminated the attribute early. Truncation happens before escaping so the 155-char budget counts reader-visible characters.
- **LF line endings.** Generated files are written with `newline='\n'`. On Windows the previous text mode wrote CRLF; git's autocrlf clean filter silently refuses to normalise any file that already contains a lone CR (one arXiv description does), so that single page would have been committed as a 581-line whole-file rewrite. Pinning LF also makes a Linux CI run and a Windows local run byte-identical.

## Coverage: before / after

| | before | after |
|---|---|---|
| HTML files in `public/` | 2,245 | 2,245 |
| pages loading the consent shim | 4 | **1,198** |
| pages with `og:title` | 34 | **1,228** |
| pages with `og:image` | 17 | **1,211** |
| pages with `twitter:card` | 17 | **1,211** |

## Verification

- Every one of the 1,194 regenerated pages parsed with `html.parser` (not regex): exactly one shim tag, always positioned before the tracker, never deferred; all nine share-card tags present exactly once; `og:title` round-trips to the exact event title; `og:url` matches the canonical; no description over 160 chars. **0 problems.**
- The regenerated diff is **head-only on 1,194 of 1,194 pages** — compared everything from `</head>` onward against `HEAD` byte for byte. No body, prose, CSS or markup change.
- `public/data/events.json` is byte-identical, confirming no pdoom-data drift snuck in with the template change.
- Spot-checked by eye: `anthropic_exodus_2021` (normal), `arxiv_2831a92843a5e489` (quoted title, correctly `&quot;`-escaped), `arxiv_407a89d50dd6ff35` (formerly multi-line description, now one collapsed line).

### Test suite

| test | result |
|---|---|
| `test-analytics-optout.js` | 17 passed, 0 failed |
| `test-design-notes.py` | 26 passed, 0 failed |
| `validate_data.py` | 0 fail, 0 warn, 10 ok |
| `test-download-resolution.js` | all checks passed |
| `generate-feeds.py --check` | OK, 14 posts |
| `check-platform-claims.py` | SKIP (see below) |
| `snapshot-copy.py --check` | 16 pages — **unchanged**, pre-existing |
| `check-stale-facts.py` | 174 findings — **unchanged**, pre-existing |
| `test_ingest_scores.py` | fails — pre-existing, stateful fixture |
| `test-header-consistency.js` | 0/25 — pre-existing, owned elsewhere |

## Found but NOT fixed

1. **1,000 orphaned `alignmentforum_*.html` pages still have no shim and no share cards.** This is the big one. The generator writes exactly 1,194 pages (the events that survive the `newsletter_archive`/`excluded` filter). The other 1,000 event pages in `public/events/` are output of an *older* run of this same template, from newsletter events the filter now drops, and pdoom-data still holds them in a separate collection `timeline_events/alignment_research/` that this sync has never read — `docs/TECH_DEBT.md` E-0, and the Chesterton's-fence case in CLAUDE.md. I deliberately did not hand-patch them: the correct fix is to make the sync read that collection, which is a separate piece of work with its own content-review implications. Mitigating fact: they are in neither `sitemap.xml` (15 URLs, no event pages at all) nor `events.json`, so they are reachable only by direct link or prior search indexing. **Full DNT honesty on event pages needs E-0 done.**
2. **`assets/og-card.jpg` is 1024x1536 — portrait.** Every page using `twitter:card=summary_large_image`, including the 4 that already did, will have it cropped hard by scrapers, which want roughly 1200x630. Not changed here since a parallel agent owns image assets; flagging so the aspect ratio gets fixed at the source rather than per page.
3. **arXiv-derived descriptions make poor share text.** They are raw PDF text extraction, so some `og:description` values read like `1 Introduction ---------------` or a run of author names. That is a data-quality issue in pdoom-data, not something the template can repair honestly — an invented summary would be worse.
4. **The `<title>` element still carries raw title text.** No event title contains `<`, so this is well-formed today, but it is one bad upstream title away from breaking. Escaping it would change 6 pages' bytes without changing a rendered character; left alone to keep this diff head-structure-only.
5. The shim adds one render-blocking same-origin request (~3.5 KB) per event page. Deferring it would remove that, and deferred scripts do execute in document order, but it would make correctness depend on the tracker tag keeping its own `defer`. Chose the robust option.

Do not merge without a look at the Netlify deploy preview for one event page.


PipFoweraker:
**Verified by curl against the deploy preview, per Pip's ruling ("verify by curl and merge"). Merging.**

Rebased onto current `main` first — which had moved twice underneath this branch (the XSS/escaper consolidation in #236, and the upstream email redaction). 13 event pages conflicted for real: `main` had redacted an author's address inside the very `<meta name="description">` this PR rewrites. Rather than hand-merge 13 pages, `sync-events.py` auto-merged cleanly (main's `redact_pii` + this PR's template work) and **all 1,194 pages were regenerated from the merged generator** against a local `pdoom-data`. The sync logged `No email addresses found in event data` — the addresses are now scrubbed upstream, so the regenerated pages keep `[email address redacted]`, and `public/data/events.json` came out **byte-identical to `main`**, confirming no upstream drift rode along.

### What curl actually returned

`GET https://deploy-preview-182--pdoom1-website-app.netlify.app/events/arxiv_f02af23f63156be9.html` → `HTTP 200, 17711 bytes, text/html; charset=UTF-8`. Head, verbatim:

```html
	<link rel="canonical" href="https://pdoom1.com/events/arxiv_f02af23f63156be9.html" />
	<meta name="description" content="1 AI Risk Skepticism Roman V. Yampolskiy Computer Science and Engineering University of Louisville [email address redacted] Abstract In this work, we surve…" />

	<!-- Share cards. Without these an event link pastes as a bare URL. -->
	<meta property="og:type" content="article" />
	<meta property="og:site_name" content="p(Doom)1" />
	<meta property="og:title" content="AI Risk Skepticism" />
	<meta property="og:description" content="1 AI Risk Skepticism Roman V. Yampolskiy … [email address redacted] Abstract In this work, we surve…" />
	<meta property="og:url" content="https://pdoom1.com/events/arxiv_f02af23f63156be9.html" />
	<meta property="og:image" content="https://pdoom1.com/assets/og-card.jpg" />
	<meta name="twitter:card" content="summary_large_image" />
	<meta name="twitter:title" content="AI Risk Skepticism" />
	<meta name="twitter:description" content="1 AI Risk Skepticism … Abstract In this work, we surve…" />
	<!-- twitter:site intentionally omitted until the handle is finalized,
	     matching public/index.html. -->

	<!-- Analytics consent shim. MUST stay above the deferred tracker below: … -->
	<script src="/assets/js/analytics.js"></script>

	<!-- Plausible Analytics -->
	<script defer data-domain="pdoom1.com" src="https://analytics.pdoom1.com/js/script.file-downloads.outbound-links.pageview-props.tagged-events.js"></script>
```

**(a) The consent shim.** Present, `src="/assets/js/analytics.js"`, **not** deferred, and physically **above** the `defer`red Plausible tag — the ordering the whole mechanism depends on. `GET /assets/js/analytics.js` → `HTTP 200, 3878 bytes, application/javascript; charset=UTF-8`, and `grep -n "createElement\|document.write\|appendChild"` over the fetched body returns **nothing** — it still creates no `<script>`, so there is no second tracker to race the first.

**(b) The share cards.** All nine tags present exactly once, `og:url`/`og:image` absolute (a relative OG URL is silently dropped by every scraper). `GET /assets/og-card.jpg` → `HTTP 200, 93595 bytes, image/jpeg` — that is the new 1200x630 card from #186, so item 2 of this PR's "found but not fixed" list is now closed rather than outstanding.

Escaping, on the event whose title contains double quotes (`GET /events/arxiv_2831a92843a5e489.html` → 200):

```html
	<meta property="og:title" content="&quot;Why Should I Trust You?&quot;: Explaining the Predictions of Any Classifier" />
```

The attribute would otherwise have terminated early on the first `"`.

**(c) Nothing else regressed.** Re-ran the PR's own structural check over the regenerated tree with `html.parser`, not regex: **1,194 / 1,194** pages carry exactly one non-deferred shim before the tracker and all nine card tags; no description exceeds 160 chars or spans lines; **0 pages** differ from `main` anywhere from `</head>` onward — the change is head-only, as claimed. Canonical, Plausible tag and `site.css` link are untouched.

### Still true, still not fixed here

The **1,000 `alignmentforum_*` pages** have no shim and no share cards. `sync-events.py` writes 1,194 pages; those 1,000 come from `timeline_events/alignment_research/` in pdoom-data, which this sync has never read (`docs/TECH_DEBT.md` E-0). Full DNT honesty on event pages still needs E-0. That gap is now tracked as its own issue alongside #224's.

Two hand-written helpers in `public/events/` (`suggest-metadata.html`, `suggest-quote.html`) are not generator output and are likewise unchanged.

PipFoweraker:
**Update after a second rebase — `main` moved again mid-merge (#234 landed).**

The previous comment's curl evidence still describes the shim and the share cards accurately, but the diff shape changed and that is worth stating plainly rather than letting the earlier "head-only" claim stand.

### What #234 changed underneath this branch

#234 added an escaper to `sync-events.py` — `esc()` / `escape_event_for_html()`, whole-record, mirroring `redact_pii()` — but **did not regenerate the pages**. So the fix was in the generator and not in the published output. Regenerating here carries it into all 1,194 pages.

Conflict resolution, as a union rather than a pick: this branch's `meta_text()` now **calls `esc()`** instead of `html.escape(..., quote=True)`, so there is still exactly **one** escaper in the module, per the rule #234 added to `CLAUDE.md`. `meta_text` keeps the two jobs `esc` does not do — collapse whitespace, truncate to 155 — and is now documented as having to run on `raw[...]`, not the pre-escaped `event[...]`, because collapse-and-truncate must happen *before* escaping or a cut can land inside an entity. `import html` went with it. `og:url` and the canonical now share `event_id_esc`, so they cannot drift apart.

`python scripts/test-sync-events.py` (#234's own suite, 55 assertions) passes, including the two shipped-corpus regressions it pins.

### The diff is no longer head-only, and here is the proof it is still safe

Measured across every page, comparing this branch to `main` from `</head>` onward:

```
bodies byte-identical to main:        1003
bodies differing ONLY by escaping:    1194
bodies differing after unescape:         0
```

Zero. Every body change is entity encoding and nothing else — `html.unescape(main_body) == html.unescape(this_body)` for all 1,194. No reader-visible text moved. The 1,003 untouched are the 1,000 `alignmentforum_*` E-0 pages plus `index.html` and the two hand-written `suggest-*.html` helpers.

Structural re-check with `html.parser`: **1,194 / 1,194** carry exactly one non-deferred consent shim positioned before the deferred tracker, all nine share-card tags exactly once, no description over 160 chars or spanning lines, and no redacted address regressed.

### Curl, against the rebuilt preview

`GET /events/arxiv_73643a60bb86bf2f.html` → `HTTP 200, 17942 bytes`. This is one of the two events #234 named as **live breakage** — its description contains `<<number to be assigned>>`, which a browser parses as a tag start both in the body and inside the `content="..."` attribute:

```html
	<meta name="description" content="Unite Paper 2021 &lt;&lt;number to be assigned&gt;&gt; A Framework for Ethical AI at the United Nations Prepared by: Lambert Hogenhout Organization: UN Office for Info…" />
	<meta property="og:title" content="A Framework for Ethical AI at the United Nations" />
	<meta property="og:image" content="https://pdoom1.com/assets/og-card.jpg" />
	<meta name="twitter:card" content="summary_large_image" />
	…
	<script src="/assets/js/analytics.js"></script>
```

Escaped, single-line, share cards present, shim above the tracker. `GET /events/arxiv_f02af23f63156be9.html` → `HTTP 200, 17715 bytes`, same shape, with the upstream `[email address redacted]` intact.

Merging.

PipFoweraker:
Follow-up on the 1,000 `alignmentforum_*` pages that still have no consent shim and no share cards: tracked in **#248**, together with #224's half of the same gap. Both are fixed by the same change — extending `sync-events.py` to read pdoom-data's `timeline_events/alignment_research/` collection (TECH_DEBT E-0).

PR #186, opened 2026-07-28
perf(payload): true 1200x630 OG card; one shared deploy-exclude list, enforced

### LINKED ISSUE #16: Provide final OpenGraph image (1200x630) and update tags
- Create/export 1200x630 image\n- Place under public/assets/ (e.g., og-1200x630.png)\n- Update og:image and twitter image\n- Add alt text if used by platform

PipFoweraker:
Payload work for TECH_DEBT **A5, B5, B6, B7**. Three of the four had already
landed on `main` and the register had not caught up — so the first thing here is
a measurement pass, and the second is finishing the two that were only half done.

**DO NOT MERGE** — for Pip's review.

## Before / after

| # | Item | Before | After | Saved | Where |
|---|---|---:|---:|---:|---|
| A5 | dashboard cat | 3,780,000 B PNG | **55,840 B** WebP 960x900 | −98.5% | already on `main` (`faa3ff7e`) — verified live, not re-done |
| B5 | `og:image` original | 2,611,379 B, 1024x1536 | *(source, kept)* | — | `/press/` links it as the press-kit logo download |
| B5 | `og-card.jpg` | 235,016 B, **1024x1536 portrait** | **93,595 B, 1200x630** | −60.2% *and the right shape* | this PR |
| B6 | `assets/dump` | shipped | excluded | 11,152,471 B (12 files) | already excluded; now in the shared list |
| B6 | `assets/image-processing-systems` | shipped | excluded | 11,615,856 B (132 files) | already excluded; now in the shared list |
| B6 | `assets/screenshots/*.png` | shipped | excluded | **9,184,720 B** (4 files) | **this PR** |
| B6 | `assets/8-bit-effect.gif` | shipped | excluded | **1,656,220 B** | **this PR** |
| B6 | *deploy total excluded* | — | — | **33,609,267 B / 149 files** | |
| B7 | `data/events.json` on the wire | 1,198,304 B | **164,755 B** gzip | −86.2% | already on `main` (`43c41b4c`) — verified live |

New in this PR: **10,840,940 B** off every deploy, **141,421 B** off the share card.

## What was actually wrong with B5

`3aa21ba6` moved all 17 `og:image` tags off the 2.49 MB painting onto a 235 KB
card — but kept the source's **1024x1536 portrait** shape. Every page declares
`twitter:card=summary_large_image`, which renders at ~1.91:1, so scrapers
centre-cropped to the cat's chest and the hand and cut the laser eyes — the
whole point of the image — out of frame. Smaller, still wrong.

`og-card.jpg` is now a deliberate 1200x630 crop around the head, produced by
`scripts/make-og-card.py` (crop box is hand-chosen and documented; `--check`
asserts 1200x630 and a 300 KB budget). **Same filename**, so no page needed
editing and no third variant exists. Closes #16.

*Scraper caches:* Facebook and X cache OG images per URL. Because the filename
is unchanged, previously-scraped shares may show the old crop for up to ~7 days.
Force a refresh via the X Card Validator / FB Sharing Debugger if that matters.

## The B6 finding that was not in the register

The excludes lived only in `auto-deploy-on-push.yml`. **Four** workflows rsync
`public/` to DreamHost — `deploy-dreamhost.yml`, `version-aware-deploy.yml` and
`weekly-deployment.yml` had none. Dispatching any of those three would have
re-uploaded every byte the first one had just stopped shipping, and reported a
clean deploy.

So the list moved to `deploy-excludes.txt` at the repo root and all four now
read it via `--exclude-from`. `scripts/check-deploy-excludes.py` enforces both
halves:

1. every workflow that rsyncs `public/` to `$DH_PATH` passes the shared list;
2. no deployed `.html`/`.css`/`.js` references a file the list drops.

Both failure paths were negative-tested (adding `assets/cats-gallery` to the
list → 6 violations in `cats/index.html`; stripping `--exclude-from` from one
workflow → 3/4 FAIL). It also rejects a CRLF filter file: rsync keeps the
trailing CR, so every pattern would match nothing — silently, and invisibly from
Windows where the whole repo is CRLF locally. `.gitattributes` pins `eol=lf` for
`deploy-excludes.txt` and nothing else.

**Why a script and not just a preview check:** Netlify serves `public/` whole,
without these excludes. A PR preview structurally cannot catch a bad exclude.

## Honest limit — one thing only Pip can do

With `--delete`, an rsync exclude **protects** the remote copy rather than
removing it. pdoom1.com still serves `/assets/8-bit-effect.gif` (1,656,220 bytes,
checked 2026-07-28). This PR stops *new* leakage; unpublishing what earlier
deploys already uploaded takes one manual command:

```
ssh $DH_USER@$DH_HOST "cd $DH_PATH && rm -rf assets/dump \
  assets/image-processing-systems assets/8-bit-effect.gif \
  assets/screenshots/*.png"
```

Left manual on purpose. Adding `--delete-excluded` to the automatic deploy would
do it, but this repo already has one story about an unattended `rsync --delete`
to production behind a guard that read as safe — that flag should be Pip's call,
not an agent's. B6 is marked **NEEDS-PIP** for exactly this.

## Verified, not assumed

- `curl -H 'Accept-Encoding: gzip' https://pdoom1.com/data/events.json` →
  `Content-Encoding: gzip`, 1,198,304 → 164,755 B, `Cache-Control: max-age=300`,
  plus the three security headers. **B7 is live and working.**
- `/assets/pdoom1-office-cat-default.webp` → 200, 55,840 B. **A5 is live.**
  `small-doom-cat.png` is a different cat and was not substituted.
- 8-bit-effect.gif and the four screenshot masters: referenced by **zero**
  deployed html/css/js across 2,250 scanned files. The masters are inputs to
  `optimize-screenshots.py` / `optimize-hero-responsive.py` and stay in git;
  every screenshot the site shows is a derived `-web.jpg` / `-web.webp` /
  `hero-bg-<w>` variant.
- `assets/dump` and `assets/image-processing-systems` are referenced only by
  `.md` READMEs documenting the pipeline — prose about the source, not a request
  a browser makes.
- `public/.htaccess` is a dotfile but `--exclude='.git*'` does not match it, so
  it does deploy. Confirmed by the live response headers above.

## Declined to touch

- **`pdoom_logo_1.png` (2.49 MB)** — `press/index.html:448` links it as the
  press-kit logo download. That is a deliberate click, not a page-load cost.
- **`og:image:width`/`height`/`alt` meta** — would touch 17 pages for marginal
  first-share gain, and `alt` would add reader-facing prose Pip has not reviewed.
  Cheap follow-up if wanted.
- **`--delete-excluded`** — see above.
- **`robots.txt`, `generate-sitemap.js`, `sync-events.py`, `public/league/`,
  `public/players/`** — owned by other agents this session.
- **`public/data/integration-health.json`** — `validate_data.py` rewrites its
  timestamp as a side effect of running. Reverted, not committed.

## Tests

Must-pass, all green: `test-design-notes.py`, `test-analytics-optout.js`,
`validate_data.py`, `check-platform-claims.py`, `test-download-resolution.js`,
`generate-feeds.py --check`, and the two new checks.

Known-failing before this branch and still failing, untouched:
`test_ingest_scores.py`, `check-stale-facts.py`, `test-header-consistency.js`.

`snapshot-copy.py --check` reports **the same 16 pages** of pre-existing drift —
no HTML was modified.


PipFoweraker:
**HOLD until after the league board opens (Fri 2026-07-31 ~1700 AEST)** — then merge.

Not a quality objection; the work is good. It rewrites **how the site ships** — four
deploy workflows switch to `--exclude-from=deploy-excludes.txt` — and the night before a
league is the wrong time to change deploy behaviour. If an exclude is wrong, the failure
mode is a file silently missing from production, discovered by a player rather than by
CI. Waiting a day costs nothing.

While it waits, one thing to fold in — from the rigour pass we started 2026-07-30.

## `check-deploy-excludes.py` is a guard with no forced-failure test

The script is the right idea and its header is honest about why it exists ("Netlify PR
previews serve public/ whole and will NOT catch a bad exclude"). But **nothing proves it
catches anything.** It has no test, so today we know it runs and exits 0 — which is
equally consistent with "the excludes are safe" and "the matcher never matches".

A guard that has only ever been observed passing has not been shown to work. Please add
a test that **forces the failure**:

- add a temporary exclude for a path that a deployed page genuinely references, and
  assert the checker **exits non-zero and names that file**;
- assert a *safe* exclude still exits 0, so the test can distinguish the two rather than
  just proving it can fail;
- exercise `matches()` against the pattern shapes actually in the file — bare name
  (`node_modules`), glob (`assets/screenshots/*.png`), and directory prefix
  (`assets/dump`) — since a matcher that silently handles only one shape would pass
  today and drop the other two from the guard's coverage without saying so.

## Worth an explicit test: the `--delete` interaction

`deploy-excludes.txt` documents something subtle and easy to get backwards:

> an exclude also PROTECTS a path from deletion, so copies already on the server persist
> and stay publicly fetchable. This list stops NEW leakage only.

That is a correctness-relevant asymmetry — "excluded" does **not** mean "removed from
production". Right now it lives only in a comment. One assertion that the tooling (or at
minimum the checker's output wording) reflects it would stop a future reader adding a
path here believing it reclaims something already served. That belief would be wrong and
would read as a privacy fix that isn't one.

## Nothing to change in the OG-card work

The 235 KB → 94 KB card and `make-og-card.py` are self-evidently verifiable and carry no
runtime failure path. No extra rigour asked there.

PipFoweraker:
**Rebased onto current `main`, CI wiring added, merging per Pip's ruling.**

The "DO NOT MERGE -- for Pip's review" line above is now stale: that review happened and the answer was yes.

### What changed while landing this

**1. Rebase conflicts in `CLAUDE.md` and `docs/TECH_DEBT.md`, resolved as a union.** `main` had moved on the same rows. Taken row by row, newest-true wins:
- `TECH_DEBT` A5 -> this branch's **DONE** (main still said OPEN); A6/A7 -> main's **FIXED**/**CLOSED**.
- B5/B6/B7 -> this branch's **DONE**/**NEEDS-PIP** rows; B8/B8a/B9 -> main's newer rows.
- The pre-PR suite list in `CLAUDE.md` kept main's CI-bracketed form, with this PR's two new checks appended.

**2. The two new checks are now actually wired to CI**, not just listed. `CLAUDE.md` on `main` gained a rule between this PR being opened and being merged -- *"When adding a check, wire it in the same commit"* -- because an audit found 10 documented checks running in no workflow at all. So `check-deploy-excludes.py` and `make-og-card.py --check` are blocking steps in `content-honesty.yml`, with `deploy-excludes.txt`, the four rsync workflows and `og-card.jpg` added to its path triggers. Both pass.

**3. That wiring turned the workflow red on a pre-existing leak, fixed in `183af55c`.** Adding the path triggers made `content-honesty` run on this PR for the first time, and `check-published-emails.py` failed on something this PR does not touch:

```
public/data/issues-cache.json: noreply@anthropic.com
FAIL. Fix at the generator where one owns the page
```

`update-game-data.yml` dumps the GitHub issues API response verbatim into that file, which `/issues/` serves; issue bodies are arbitrary third-party text and one carried a `Co-Authored-By` trailer. Fixed at the generator (the workflow now applies `sync-events.py`'s `redact_pii` to the payload -- reused, not re-implemented, so there stays exactly one definition of "what an email address looks like"), and the committed file scrubbed with `--fix` so it matches what the generator will now produce. Without the generator half, the 6-hourly cron would have put the address straight back.

That guard was silently failing on `main` whenever it ran. Worth knowing separately from this PR.

### Unchanged and still true

The **NEEDS-PIP** item stands: with `--delete`, an rsync exclude *protects* the remote copy rather than removing it, so pdoom1.com still serves the ~34 MB earlier deploys already uploaded. Reclaiming it is one manual `ssh ... rm -rf`, deliberately not automated -- the command is in the `auto-deploy-on-push.yml` comment and in the PR body above.

PipFoweraker:
**Correction to the comment above.** The `issues-cache.json` email leak I described was fixed on `main` in parallel (#234's lane) while this branch was rebasing, with a better version of the same fix — it counts and logs what it redacted. My duplicate commit has been dropped rather than merged on top; `check-published-emails.py` now passes on this branch against current `main` with no change from here.

Everything else in that comment stands: the union conflict resolution in `CLAUDE.md` / `TECH_DEBT.md`, and the two new checks being wired as blocking steps in `content-honesty.yml`.

Issue #177, opened 2026-07-26
stats: uplift audit for calculate-game-stats once real data channels land


PipFoweraker:
Capturing the stats-uplift opportunity so it's ready the moment the data channels land.

## Context

`scripts/calculate-game-stats.py` (this repo's) produces three figures, all currently stubbed:
- `frontier_labs_count` — counts a **hardcoded list** in the script.
- `baseline_doom_percent` — literal `23`.
- `strategic_possibilities` — literal `10000`.

Two data channels are now in motion that let us retire these stubs:
- **Frontier labs:** pdoom-data#37 → pdoom1#962 will expose a fetchable list → drive `frontier_labs_count` from it (and delete the hardcoded list).
- **p(Doom) baseline + "possibilities":** should come from the game's **calibration emit** (baseline p(Doom), the sim's 100%-topout date) — a pending pdoom1 ask.

## Do (when the data lands)

1. Wire `frontier_labs_count` to the pdoom1-exposed frontier-labs list; remove the hardcoded list.
2. Wire `baseline_doom_percent` + `strategic_possibilities` to the game's calibration artifact; remove the literals.
3. **Drive-by uplifts** now that the engine has advanced and we're starting to collect game data:
   - New stat: **install base** (from the #942 anonymous launch-ping via Plausible).
   - New stat: **downloads by channel** (already flowing via the UTM-stamped Download event).
   - Audit the engine for any other facts cheaply + honestly surfaceable as stats.
4. Hold the fail-hidden discipline throughout: a dash on fetch failure, never a stale literal.

**Blocked on:** pdoom-data#37, pdoom1#962, and the p(Doom) calibration emit.

PipFoweraker:
Blocker chain now fully lodged — the third dependency has a number.

- **pdoom1#994** — `calibration.json` artifact: baseline p(Doom) (as a 0..1 fraction with an explicit unit field), the sim's 100%-topout date, and the ladder/patch version, attached to every release as `releases/latest/download/calibration.json`. Retires `baseline_doom_percent: 23`. It also carries a `game_stats.strategic_possibilities` block with a required `_definition` string, which retires `strategic_possibilities: 10000` — a smaller true number beats a larger impressive one.
- **pdoom1#962** / **pdoom-data#37** — frontier-labs list. Retires `frontier_labs_count` and the hardcoded lab-name list plus its `max(n, 5)` floor.

[...]

Issue #191, opened 2026-07-28
League week scope line: freeze the score-API / board-key / download-pointer surface after Thu 2026-07-30


PipFoweraker:
Heads-up for everyone working in this repo this week, including parallel agents. **This is a narrow scope line, not a freeze.** Most of the repo is completely unaffected.

## Why

P(Doom)1 opens a new league **Fri 2026-07-31 ~1700 AEST** on ladder epoch **L3** (see #151 for the exact board-key strings). The game side has stage gates that prove the shipped artifact, but those gates only prove OUR artifact. One of them -- "board-key fork verified clean" -- **spans both repos** and cannot be verified from the game side alone.

If the score-API surface, the board-key handling, or the download pointers move between now and Friday evening, the league can break in a way none of the game-side gates can see.

## The line

After **Thu 2026-07-30**, please do not change, without a heads-up on this issue:

- the **score API surface** -- GET/POST shape, the `version` and `seed` params, auth/token handling
- **board-key handling** -- anything that buckets, filters, renames or migrates boards
- **download pointers** -- whatever resolves "latest" / the download button to a release artifact

## What is explicitly still open season

Everything else. Content, copy, tone passes, page layout, navigation, the forum work, the grant-readiness backlog, CI repair, docs. None of it can break a league. Do not slow down on any of that.

## Related open items in this repo (context, not asks)

- #151 -- the ladder epoch cut. Just updated with the L3 correction and the exact board-key strings. **Item 1 in that comment is the only Friday-blocking one.**
- #139 -- site still resolves version from hardcoded `0.4.1` / `0.6.0`. Worth knowing during league week, since a bump will not propagate to what players see.
- #147 -- download pointer currency.
- #126 -- weekly-league rollover false-positive concern. Friday IS a rollover, so this one is live rather than theoretical.

## After Friday

The line lifts once the board is open and first runs are observed. Saturday is a hotpatch watch window on the game side; if something on the website side needs to move Friday evening, say so here and we will coordinate rather than block.

Filed from the pdoom1 side during league-week prep, 2026-07-29.

PipFoweraker:
Closing: the freeze window it describes has passed.

The league opened Friday 2026-07-31 on `(weekly-2026-w31, L3)`, the seed was blessed that evening and recorded in `ladder-epochs.json` (PR #237), and the board is live with real entries. `week_info` on `main` reads `2026_W32`, 07-31 → 08-06, status running.

[...]

PR #196, opened 2026-07-29
feat(frontier-labs): pin the definition, derive the count, stamp what is pending

PipFoweraker:
Pins the frontier-labs definition, discussion and tracking-over-time to `/frontier-labs/`, derives the count from data instead of a hardcoded list, and introduces the reusable under-construction motif other pages will adopt.

**Do not merge yet** — Pip to review the copy and the stamp direction first.

## The number was wrong three different ways

The page published **7** frontier labs. `public/data/version.json` published **5**. `scripts/update-version-info.py` asserted **7**. The 6-hourly `update-game-data` workflow wrote **5**. The page itself listed six real labs.

Tracing it back: `calculate-game-stats.py` held a list of ~20 lab names, substring-searched `public/index.html` for each, and returned `max(len(matches), 5)` — with a bare `return 7` if the homepage was missing.

- The list was invisible to readers, so the number was unfalsifiable.
- The floor meant the published figure had settled on exactly `5` — the floor, not a count.
- The `return 7` fallback shipped precisely when the real lookup failed.
- And the page's `7` was six real labs **plus one hypothetical entrant**, summed. A modelling parameter is not an organisation.

## What this does

**`public/data/frontier-labs.json` is now the single source of truth.** The count is `roster.labs where kind == "real" and status == "active"` — the same predicate evaluated by `calculate-game-stats.py` and by the page in the browser, so the two structurally cannot disagree. No floor, no fallback literal: if the roster is unreadable the script raises and the previous value stands.

**Three writers were silently competing for `frontier_labs_count`.** `update-game-data.yml` and `update-version-info.py` both restated literals and would have reverted the derived value within six hours. Both now carry `game_stats` forward instead, and the workflow runs the derive step.

**The definition and the discussion are pinned as static prose** in the page source — four criteria, and six things genuinely contested about them. They survive without JavaScript and change only in a reviewable diff. Deliberately *not* duplicated into the JSON: two copies of an argument drift apart, and the one a reader sees would be the one nobody edited.

**Tracking over time renders as stamped "awaiting data".** No such series exists in any repo — a snapshot of founding years cannot say when a lab *stopped* being one. So there is no chart. The artifact URL and schema this page will consume are written into the roster file under `tracking`, so the implementing repo does not re-derive them.

**Removed the per-lab "Compute" and "Safety Focus" ordinals** (Very High / High / Medium). Invented scales with no source, sitting beside sourced founding years and looking equally solid. The removal is explained on the page.

**The roster is known-incomplete and now says so in data.** `known_omissions` is a first-class list — DeepSeek, Alibaba (Qwen), Moonshot AI, Zhipu AI, and others that meet the definition but have no sourced row. The first four are missing purely because the original list was Anglophone; that is a defect in the roster, not a judgement about the labs.

## The under-construction motif

`public/css/stamp.css` — a diagonal bureaucratic rubber stamp. Mid-2000s office paperwork, not a Web 1.0 GIF.

Pure CSS, no images, no external fonts, no network requests. **Opt-in per page** via one `<link>`, deliberately *not* in `site.css` (that file loads nearly everywhere and has whited out the site once already). All classes namespaced `.stamp*`, so cascade order does not matter.

Four cheap layers: `rotate(-7deg)` for the slam angle; a `::before` duplicate border box at 1–3px offset for the misregistered second impression; `text-shadow` for the same on the letterforms; a `::after` angled `repeating-linear-gradient` in the page background colour with uneven stops for streaky ink voids.

- **Nothing moves.** A stamp is a static impression. The `prefers-reduced-motion` block only neutralises transitions a host page might inherit onto it.
- **Stamp words are real text** with a visually-hidden `Status stamp:` prefix, not `content:` on a pseudo-element — they carry meaning about how far to trust the data underneath.
- **Ink weathering is an overlay, not a mask.** A mask would look better; its failure mode is an *invisible* stamp, which is the one thing a trust label cannot do. An overlay's worst case is a clean stamp. Degrade toward more honest, not less.
- **It is a label, not a censor bar.** Default placement is in normal flow. `.stamp--overlay` exists for panels that are themselves pending, is `pointer-events:none`, and falls back into flow below 480px.
- **No emoji.** It is typography.

Variants are one property each (`--stamp-ink` + angle): `--awaiting`, `--provisional`, `--review`, `--incomplete`, `--restricted`. **Pick the one that is true of the block** — a stamp saying something false launders a guess into an official-looking record.

Recipe, adoption snippet and degradation checklist: `docs/UNDER_CONSTRUCTION_STAMP.md`.

## Cross-repo

Both existing issues were extended rather than duplicated, each with the exact artifact URL, schema and precision rules:

- pdoom-data#37 — https://github.com/PipFoweraker/pdoom-data/issues/37#issuecomment-5111546704
- pdoom1#962 — https://github.com/PipFoweraker/pdoom1/issues/962#issuecomment-5111549252

Both block #177.

## Tests

Passing: `test-design-notes.py`, `test-analytics-optout.js`, `validate_data.py` (1 pre-existing league-rollover WARN), `check-platform-claims.py`, `test-download-resolution.js`, `generate-feeds.py --check`.

Known-failing and unchanged by this branch: `test_ingest_scores.py`, `check-stale-facts.py`, `test-header-consistency.js` (0/25 before and after this branch — identical error count), `snapshot-copy.py --check` (the same 35 pages are flagged before and after; this branch adds none).

Prose changed on: `public/frontier-labs/index.html` (substantial rewrite), `public/game-stats/index.html` + `public/stats/index.html` (the Frontier Labs stat description, which described the old mention-counting method and is now false), `public/index.html` (a hardcoded `5` placeholder → `--`, matching its two sibling stats).

### DIFF
--- a/public/data/frontier-labs.json
+++ b/public/data/frontier-labs.json
@@ -1,0 +1,166 @@
+{
+  "_README": "Single source of truth for the frontier-labs roster on /frontier-labs/ and for game_stats.frontier_labs_count. NOTHING downstream may hardcode a lab list or a lab count: scripts/calculate-game-stats.py derives the count from this file and fails loudly if the file is missing, rather than substituting a literal. The roster itself is INTERIM and hand-entered; its destination is pdoom-data (issue #37), consumed via pdoom1#962. The inclusion boundary is prose, pinned at https://pdoom1.com/frontier-labs/#definition, and is versioned by definition_version so a later boundary change is visible as a change rather than a silent recount.",
+  "version": 1,
+  "definition_version": 1,
+  "definition_ref": "https://pdoom1.com/frontier-labs/#definition",
+  "definition_status": "under_review",
+  "as_of": "2026-07-29",
+
+  "stat_contract": {
+    "frontier_labs_count": "COUNT of roster.labs where kind == 'real' AND status == 'active'. It is the size of a published, interrogable roster -- NOT a claim about how many frontier labs exist in the world. The roster is known-incomplete; see known_omissions.",
+    "excluded_from_count": "hypothetical entries. A modelling parameter is not an organisation, and adding one to a list of real ones is a category error. The page's previous headline figure of 7 was 6 real labs + 1 hypothetical entrant summed together.",
+    "no_floor": "There is no minimum. An earlier implementation returned max(count, 5), so the published number was the floor and not a count."
+  },
+
+  "roster": {
+    "status": "provisional",
+    "completeness": "known_incomplete",
+    "source": {
+      "provider": "hand-entered",
+      "upstream": "pdoom-data#37",
+      "status": "interim",
+      "note": "Carried forward from the prose that was already on /frontier-labs/. Founding years are recorded at year precision because that is the precision we can stand behind; per-row primary citations are part of pdoom-data#37's deliverable."
+    },
+    "source_quality": "tertiary",
+    "labs": [
+      {
+        "id": "openai",
+        "name": "OpenAI",
+        "kind": "real",
+        "status": "active",
+        "founded": "2015",
+        "founded_precision": "year",
+        "admitted_under": 1,
+        "url": "https://openai.com",
+        "reference": "https://en.wikipedia.org/wiki/OpenAI",
+        "note": "Trains its own foundation models at the largest publicly known scale."
+      },
+      {
+        "id": "anthropic",
+        "name": "Anthropic",
+        "kind": "real",
+        "status": "active",
+        "founded": "2021",
+        "founded_precision": "year",
+        "admitted_under": 1,
+        "url": "https://anthropic.com",
+        "reference": "https://en.wikipedia.org/wiki/Anthropic",
+        "note": "Founded by former OpenAI researchers; trains the Claude models."
+      },
+      {
+        "id": "google-deepmind",
+        "name": "Google DeepMind",
+        "kind": "real",
+        "status": "active",
+        "founded": "2010",
+        "founded_precision": "year",
+        "admitted_under": 1,
+        "url": "https://deepmind.google",
+        "reference": "https://en.wikipedia.org/wiki/Google_DeepMind",
+        "note": "Founded 2010 as DeepMind; acquired by Google 2014; merged with Google Brain in 2023 to form Google DeepMind. Counted as one entity under criterion 3 -- it sets its own research agenda. Reasonable people count the lineage differently."
+      },
+      {
+        "id": "meta-ai",
+        "name": "Meta AI",
+        "kind": "real",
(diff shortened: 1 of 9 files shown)


PipFoweraker:
**HOLD until after the league board opens (Fri 2026-07-31 ~1700 AEST)** — then merge.

Not a quality objection. This edits `public/data/version.json`, which auto-updates when
v0.13.2 is released tomorrow. Merging tonight buys a guaranteed conflict at the worst
possible moment, and `#191`'s scope line asks for a heads-up before anything touches the
download-pointer surface during league week. Nothing here is league-critical, so waiting
costs nothing and rushing could cost the release.

While it waits, two things to fold in — from the rigour pass we started 2026-07-30.

## 1. The fallback removal is right, and it is untested

The module docstring makes a strong and correct claim:

> `return 7` shipped precisely when the real lookup failed … There is no floor and no
> fallback literal. If the roster cannot be read, this fails.

**That claim currently has no test.** A safety property asserted only in a docstring is
documentation, not evidence — and this one is exactly the class CLAUDE.md singles out,
so it deserves a forced failure rather than a promise. Please add a test that:

- makes the roster unreadable (missing file / malformed JSON / empty list) and asserts
  the script **exits non-zero and writes nothing**, and
- asserts the previously-published value is left intact rather than replaced by a
  smaller/zero count — the dangerous outcome is not a crash, it is a confident wrong
  number replacing a good one.

The pattern to copy is now on `main`: `scripts/test-board-escaping.js` and the rewritten
`scripts/test_ingest_scores.py` both build inputs in temp dirs and assert the refusal
path, not just the happy path.

## 2. Two sibling fallback literals survived the sweep

`scripts/calculate-game-stats.py:132-133`:

```python
game_stats.setdefault('baseline_doom_percent', 23)
game_stats.setdefault('strategic_possibilities', 10000)
```

These are the same shape as the `return 7` this PR removes: a literal that ships
**precisely when the real lookup failed**, presented to a reader as a current fact. The
frontier-labs count got the honest treatment; `baseline_doom_percent` and
`strategic_possibilities` did not, and `23%` reads to a visitor as a measured finding
about AI risk.

Either give them the same treatment (fail, or preserve last-known-good, and stamp
`pending`), or — if they are genuinely declared constants rather than derived values —
say so in the file with a `source`, so the next reader can tell a *declaration* from a
*measurement*. The distinction is the whole game here.

Not asking for scope creep: flagging in the same PR because it is the same defect class
and this is the file where someone already has the context.

## 3. One test for the "pending" stamp

The PR's value is that it stamps what is pending rather than inventing it. Worth one
assertion that a pending stat renders **as pending** and never as a number — that is the
regression that would quietly undo the whole point.

PR #198, opened 2026-07-29
fix(encoding): kill the cp1252 trap repo-wide, and guard it so it cannot come back

PipFoweraker:
## The bug, both halves

Windows consoles default to cp1252. That produces two distinct failures, and this repo has been bitten by both.

**The loud half.** A `print()` of any non-ASCII character raises `UnicodeEncodeError` on the *first* print, killing the script before it does any work. `health-check.py` died this way for months, and the resulting traceback — which names the interpreter's own `encodings/cp1252.py` — was captured into a published JSON file and served from pdoom1.com. A Python stack trace became reader-facing content. `test-changelog-structure.py` was **still** dying on `✓` as of this branch's base, so the documented test suite was lying about its own health.

**The quiet half, which is worse.** `open(path)` with no `encoding=` decodes as cp1252 on Windows and utf-8 on Linux, so a UTF-8 file mojibakes *without raising*. On 2026-07-28 a diagnostic read a file that way, mistook a mangled em dash for data corruption, and produced a **false bug report** that cost real time to disprove. The write side stops a script; the read side makes people believe untrue things about the data. Reproduced live in this session:

```
>>> open('CHANGELOG.md', 'r').read()
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 4023
```

## Audit

**43 Python modules** in the repo. 4 excluded (`public/assets/image-processing-systems/dump/` — a vendored dump of a retired pipeline, served as static bytes, never executed). **39 audited, 61 findings, 32 modules changed.**

| | before | after |
|---|---|---|
| modules that print without the preamble | 32 | 3 (all waived, see below) |
| `open()` in text mode without `encoding=` | 19 | 1 (waived) |
| `subprocess` in text mode without `encoding=` | 9 | 0 |

Changes: 29 preamble insertions (8 also needed `import sys`), 18 `open()` calls, 9 subprocess calls, 4 `open()` calls inside workflow heredocs. No script's behaviour changed — only how it encodes I/O.

## Structural decision: duplicate the preamble, do not share it

I weighed a `scripts/_encoding.py` helper and rejected it.

These scripts are invoked directly (`python scripts/foo.py`) from many working directories, and from workflow steps that `cd` elsewhere. A helper import therefore needs `sys.path` surgery *above* it — which is more lines than the preamble itself, and which can fail. An import that raises defeats the entire purpose of a preamble whose one job is to survive being the first thing that runs. CLAUDE.md already prescribes this literal snippet, so a divergent mechanism would make the documented one wrong.

So: the same 8-line block, verbatim, in every printing module, with the checker enforcing it.

One deviation from a pure copy-paste: the `subprocess` fixes use `encoding="utf-8", errors="replace"`, not bare `encoding="utf-8"`. Strict decoding would turn a child process emitting odd bytes into a *crash* where it previously mojibaked through — that is a behaviour regression on the health-check path. `open()` on the read side is left strict on purpose: a repo data file that is not valid UTF-8 is genuinely broken and should say so.

## Every `open()` without an encoding, on the reading side

The dangerous ones (read `'r'` or default mode — these silently produced wrong bytes):

| file:line | what it was reading |
|---|---|
| `scripts/calculate-game-stats.py:80` | `public/data/version.json` |
| `scripts/check-version-deployment.py:17` | `package.json` |
| `scripts/check-version-deployment.py:138` | `public/data/website-changes.json` |
| `scripts/check-version-deployment.py:148` | `public/blog/index.json` |
| `scripts/check-version-deployment.py:159` | `public/data/status.json` |
| `scripts/prepare-weekly-deployment.py:121` | `package.json` |
| `scripts/test-changelog-structure.py:37` | `public/data/{website,game}-changes.json` |
| `scripts/verify-deployment.py:65` | every JSON file it validates |
| `scripts/verify-deployment.py:129` | `public/data/version.json` |
| `.github/workflows/extract-analytics.yml:110` | each analytics JSON |
| `scripts/sync/sync-events.py:952` | **not fixed — see waivers** |

Write-side `open(..., 'w')` without encoding (would have written cp1252 bytes into files the site serves as UTF-8): `calculate-game-stats.py:107`, `extract_analytics.py:289`, `prepare-weekly-deployment.py:309`, `test-orchestrator.py:219`, `update-version-info.py:164`, `verify-deployment.py:223`, `extract-analytics.yml:134`, `update-game-data.yml:103` and `:123` (the last two write `public/data/version.json` and `issues-cache.json` — files served to visitors).

`subprocess` text-mode decodes with no encoding, all now `utf-8`/`replace`: `ab-preview.py:41`, `check-version-deployment.py:28`, `extract_analytics.py:82`, `health-check.py:123`, `prepare-weekly-deployment.py:53`, `test-integration.py:72,111,220`, `test-orchestrator.py:41`, `test_ingest_scores.py:28`.

## The guard

`scripts/check-encoding-safety.py` — AST-based, not regex-based, so `"open("` inside a docstring or an HTML template string does not trip it. It checks:

- **W1** a module that can print but lacks the preamble. Any `print()` counts; the argument does not have to be a literal emoji, because filenames, JSON values, HTML prose and `repr()`'d exceptions are all routes for a non-ASCII byte to reach stdout.
- **R1/R2/R3** `open()` / `Path.read_text|write_text` / text-mode `subprocess` without an explicit encoding.

Writing it caught a bug in itself worth flagging: the first version matched on the *last* attribute name, so `Image.open(path)` and `webbrowser.open(url)` were "fixed" with an `encoding=` kwarg neither accepts — which would have crashed `optimize-screenshots.py` at runtime. `_is_builtin_open()` now requires `ast.Name` or `io`/`codecs`, and both shapes are pinned in the self-test.

**Proof the guard fails.** Copied `test-changelog-structure.py` to a scratch file, stripped the preamble and restored the bare `open()`:

```
FAIL: 2 encoding-safety finding(s) across 40 Python modules
  scripts/_scratch-regression.py:1:  [W1] module prints but has no stdout/stderr reconfigure preamble
  scripts/_scratch-regression.py:43: [R1] open() in text mode without encoding=
GUARD EXIT=1
```

Running that same scratch copy under `PYTHONIOENCODING=cp1252` reproduces the original crash exactly, and the fixed file passes:

```
--- BEFORE ---
Testing changelog structure...
  File "...\_scratch-regression.py", line 35, in test_changelog_files
    print(f"✓ {name} exists")
  File "...\Lib\encodings\cp1252.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode character '✓'
exit=1
--- AFTER ---
✓ All changelog tests passed!
exit=0
```

Scratch file deleted; guard back to exit 0.

`--self-test` (19 cases) is a permanent version of that proof, wired into CI so the checker is never green because it stopped checking.

## CI

New `encoding-safety.yml`, two jobs. The ubuntu job runs the static sweep and the self-test. **The second job runs on `windows-latest` with `PYTHONIOENCODING=cp1252`** and executes the real scripts, because ubuntu CI structurally cannot observe either symptom — a green ubuntu check is no evidence at all that this bug is gone. Read-only, no secrets, opens no issues.

## A trap worth recording (now in CLAUDE.md)

**The agent shell lies about this bug.** Claude Code sets `PYTHONIOENCODING=utf-8:surrogateescape` and code page 65001, so the print crash does **not** reproduce in-session even though Pip hits it in his own terminal. My first "before" measurement showed `test-changelog-structure.py` passing on unmodified `origin/main` — which is false for Pip. Every before/after below was re-measured under a forced cp1252 stdout. Note that `locale.getpreferredencoding()` is still cp1252 in that shell, so the *read* side misbehaves in-session regardless — which is presumably how the em-dash false alarm happened.

## Exit gate — full local suite on Windows, `PYTHONIOENCODING=cp1252`

| test | before | after |
|---|---|---|
| `test-design-notes.py` | PASS | PASS |
| `test-analytics-optout.js` | PASS | PASS |
| `test_ingest_scores.py` | FAIL | FAIL *(known, not mine)* |
| `validate_data.py` | PASS | PASS |
| `check-stale-facts.py` | FAIL | FAIL *(known, not mine — 174 findings both sides)* |
| `check-platform-claims.py` | PASS | PASS |
| `test-download-resolution.js` | PASS | PASS |
| `snapshot-copy.py --check` | FAIL | FAIL *(known, not mine — 16 pages/19 added both sides, so no prose drift introduced)* |
| `generate-feeds.py --check` | PASS | PASS |
| `test-header-consistency.js` | FAIL | FAIL *(known, not mine)* |
| **`test-changelog-structure.py`** | **FAIL** | **PASS** |
| `check-encoding-safety.py --self-test` | n/a | PASS |
| `check-encoding-safety.py --check` | n/a | PASS |

Every result other than `test-changelog-structure.py` is identical to `origin/main`.

## Files skipped for concurrent agents — FOLLOW-UP NEEDED

All three genuinely need the fix. They are waived by name in `KNOWN_UNFIXED` and **printed loudly on every guard run**, so the waiver cannot go quiet. Delete the entry as each branch merges.

- **`scripts/ingest_scores.py`** — W1. Also matters transitively: `test_ingest_scores.py` runs it as a subprocess, so under cp1252 the child can still die on its own first print.
- **`scripts/sync/sync-events.py`** — W1 **and R1 at line 952** (`open()` in text mode). This is the generator for ~2,194 event pages; reading its source data through cp1252 is exactly the mojibake path that caused the false bug report.
- **`scripts/weekly-league-manager.py`** — W1.

Non-Python paths on the do-not-touch list contain no Python and needed nothing: `public/robots.txt`, `scripts/generate-sitemap.js`, `public/assets/js/navigation.js`, `public/blog/`, `public/press/`, `public/privacy/`, `public/frontier-labs/`, `public/bug-report/`, `public/leaderboard/`, `public/league/`.

## Reported, not fixed (behaviour changes belong in their own PR)

1. **Four modules now carry two encoding mechanisms.** `bugfix_pass_20260715.py`, `validate_data.py`, `sync/merge-alignment-events.py` and `metadata/tag_newsletters.py` already had a hand-rolled `if sys.platform == 'win32': sys.stdout = io.TextIOWrapper(...)` hack. The preamble runs first and the old hack then replaces `sys.stdout` on top of it. Harmless — the second wrapper sits on the same buffer, which is what Python already does today — but the old hack is strictly worse (Windows-only, stdout-only, no `errors="replace"`) and should be deleted. Removing it is a behaviour change, so it is not in this PR.
2. **`scripts/check-encoding-safety.py` also reports `E0` for unparseable files.** A UTF-8 BOM at the start of a `.py` trips it. Nothing in the repo currently does; noting it so the code is not mistaken for dead.

## Where I could not be certain a script emits non-ASCII

Nowhere — and that is the point of the rule rather than a gap in the audit. The checker deliberately does **not** try to prove a given `print()` is ASCII-only. Every one of these scripts prints filenames, JSON values, HTML prose or `repr()`'d exceptions, all of which are routes for an em dash or an arrow to reach stdout from data this repo is full of. Proving the negative per-call-site would be unsound and unmaintainable; requiring the preamble on any module that prints costs 8 lines and cannot be wrong.

The one place a judgement call was made is the **exclusion**: the 4 files under `public/assets/image-processing-systems/dump/`. They are Python, and they do print. I excluded them because they are a vendored dump of a retired pipeline that this repo never executes, and they sit under `public/` — editing them would change bytes deployed to pdoom1.com for no behavioural gain. The exclusion is a named constant with that reason attached, so it is visible rather than silent. If that call is wrong, the fix is a one-line deletion in `EXCLUDED`.


PipFoweraker:
Rebased onto `main` (2026-07-31). Three conflicts resolved:

- **`scripts/test_ingest_scores.py`** — took **main's** version, not this branch's. `chore/hygiene` (#185) rewrote this test far more thoroughly in parallel (five cases, synthetic seed docs, no committed-fixture dependency) and it already carries the cp1252 guard. This branch had patched the older file. Nothing lost.
- **`scripts/alpha-watch.py`** — kept this branch's guard; both sides added the same fix, this one has the better comment.
- **`CLAUDE.md`** — union; the two sides documented different subsystems.

**Independent evidence this matters:** `python scripts/test-integration.py` fails on `main` today with `UnicodeEncodeError: 'charmap' codec can't encode character '\u274c'` — dying on its first non-ASCII print, before doing any work. On this branch it **passes**. That is a second currently-red check this PR turns green, on top of the class it closes.

Issue #200, opened 2026-07-29
copy accuracy: site says builds are 'free and open source' (they are not), and the footer version looks wrong


PipFoweraker:
Two factual inaccuracies found while verifying that the corrected macOS instructions from #195 went live (they did -- that fix is confirmed good on the live site, thank you).

Both matter more than usual this week: the Manifund application went live 2026-07-29 and funders are now clicking through to this site.

## 1. "All builds are free and open source" -- they are not open source

The live download page states this. The game repo's `LICENSE` says:

> P(Doom)1 -- Interim License (pre-1.0)
> Copyright (c) 2026 Pip Foweraker. All rights reserved.
> Source is provided for reference, evaluation, and contribution. No license is granted to redistribute, sell, or make commercial use of this software or its [...]

That is **source-available**, not open source. The two are not synonyms, and the audience this project is courting -- AI-safety people and the gaming community -- polices the distinction closely. Claiming an OSI-style licence the project does not have is the kind of error that gets quoted back at you.

**Suggested replacement copy:** "Free to download and play. Source-available -- you can read and contribute to the code; see the LICENSE for terms." Accurate, and keeps the genuinely good part of the claim.

This is a copy fix, not a licensing decision. If the licence itself should change, that is a separate and much slower conversation.

## 2. Footer shows v1.1.1

The current game release is **v0.13.1**. If that footer number refers to the website's own version, ignore this -- but it reads as a game version to a visitor, and a funder seeing "v1.1.1" then landing on a v0.13.1 release page will assume something is stale or wrong.

Related to #139 (version single-source-of-truth -- resolve the displayed version from a canonical source rather than hardcoded strings). If #139 lands, this class of error stops recurring.

## Not blocked by the freeze

Copy-only. Outside the #191 scope line -- no score API, board keys, or download pointers touched.

## Confirmed working, for the record

The macOS instructions are live and correct on both branches: System Settings -> Privacy & Security -> Open Anyway for Sequoia 15+, right-click -> Open for macOS 14 and earlier, with honest platform status ("Windows is the primary tested build; macOS and Linux are experimental"). Matches the game README exactly. Good, fast work.

PR #224, opened 2026-07-31
content(events): stamp the impacts block and show direction, not magnitude

PipFoweraker:
Pip ruled **variant B** on 2026-07-31 after an A/B comparison.

## Two changes, one f-string each, 1,194 pages

1. The **Game Impacts** block carries a `stamp--restricted` — *"Not verified in game"* — and a plain-English note.
2. The **Change** column becomes **Direction**: `proposed: down` rather than `-80`.

## Why the magnitude goes and the row stays

Of **1,194** events, the calculated effects reach gameplay in one branch — **7 events**. The other **1,174 are flavour**, and flavour is **hidden by default** (`_feed_important_only`). So a precise figure was being presented, on ~2,190 pages, for events that move nothing.

Direction is the part that is probably right. Magnitude is the part that is definitely unverified — and precision reads as authority regardless of the caveat above it. **A stamp over a number still shows the number, and the number was the false part.**

The magnitudes are **not deleted**. They live in the corpus, and the per-page suggestion links go straight there. Nothing is hidden; it is simply no longer asserted by us.

## Precedent, from the game's own repo

`godot/scripts/core/events.gd:290` — displayed doom numbers were removed there for exactly this reason:

> a literal "doom" effect must NEVER reach `Resources.add`'s `state.doom += v` sink — that write is CLOBBERED at resolve … so every "(+/-N doom)" message in event content **was a silent lie**.

And pdoom-data's own contract brief, line 33: *"Nothing in the candidate feed has been human-reviewed. **Do not ship it to players yet.**"*

## Known gap — deliberately not fixed here, needs a decision

**1,000 `alignmentforum_*` pages still show the old numeric impacts and carry no stamp.** They are generated from a **separate** pdoom-data collection — `timeline_events/alignment_research/` — that `sync-events.py` has **never read**. This is the Chesterton's fence CLAUDE.md warns about: they look orphaned and are in fact the only published surface for a live dataset.

So this change covers **1,194 of 2,194** event pages. The other 1,000 are now the *only* pages on the site still making the claim we just ruled against.

**What I found while checking:** the collection exists locally with a **matching schema** — 1,000 records carrying the same `impacts`, `pdoom_impact`, `rarity`, `sources` fields. So extending the sync to read it is **cheaper than TECH_DEBT E-0 assumes**.

**Why I did not just do it:** regenerating 1,000 pages from a source that has been unread for months could shift far more than the impacts block — titles, descriptions, categories, sources. That is a content change of unknown size on a tenth of the site, made while nobody is looking at it. It wants a decision and a diff review, not a unilateral regenerate.

Three options, roughly ascending in cost:

1. **Leave them**, and accept that a tenth of the event pages disagree with the other nine tenths until the sync is extended.
2. **Post-process just the impacts block** on the existing 1,000 files — makes the site consistent today, but hand-editing generated files creates exactly the un-regenerable artefact that produced this problem.
3. **Extend `sync-events.py` to read the second collection** — the real fix, closes TECH_DEBT E-0, and needs a review of what actually changes on those 1,000 pages.

My recommendation is **3**, not today.

### DIFF
--- a/docs/UNDER_CONSTRUCTION_STAMP.md
+++ b/docs/UNDER_CONSTRUCTION_STAMP.md
@@ -1,0 +1,128 @@
+# The under-construction stamp
+
+A diagonal rubber-stamp impression — the thing a clerk slams on a form.
+Mid-2000s office paperwork, not a Web 1.0 "under construction" GIF.
+
+Source: `public/css/stamp.css`. First adopter: `public/frontier-labs/index.html`.
+
+## What it is for
+
+Marking a block of the site whose data is **not yet what it will be**, without
+either (a) hiding the block or (b) letting a reader mistake a placeholder for a
+measurement. It is the visual half of the same discipline `/state-of-doom/`
+applies in text: render *"awaiting &lt;source&gt;"*, never a guessed value.
+
+It is a **label, not a censor bar**. If a stamp is covering something the reader
+needs, the stamp is wrong.
+
+## Adopting it on another page
+
+Two lines. One `<link>`, one block of markup.
+
+```html
+<!-- in <head>, AFTER the page's own inline <style> is not required —
+     every class is namespaced .stamp*, so order does not matter -->
+<link rel="stylesheet" href="/css/stamp.css">
+```
+
+```html
+<div class="stamp-block" role="note">
+  <p><span class="stamp stamp--awaiting"><span class="stamp-sr">Status stamp: </span><span class="stamp-text">Awaiting data</span></span></p>
+  <div class="stamp-body">
+    <p>What this block will show, what it is waiting on, and who is doing it.</p>
+  </div>
+  <p class="stamp-docket">
+    Form PD-1/FL &middot; Ref <a href="https://github.com/PipFoweraker/pdoom-data/issues/37">pdoom-data#37</a> &middot; Clerk: unassigned
+  </p>
+</div>
+```
+
+That is the whole component. No JS, no images, no external fonts, no network
+requests — the site's CSP posture stays intact.
+
+### Required bits, and why each one
+
+| Bit | Why it is not optional |
+| --- | --- |
+| `role="note"` on the block | tells assistive tech the panel is an aside about the content, not the content |
+| `<span class="stamp-sr">Status stamp: </span>` | a sighted reader knows it is a stamp from the border and the angle. Nobody else does. |
+| Sentence case in the HTML (`Awaiting data`) | CSS does the shouting via `text-transform`. Several screen readers spell ALL-CAPS source text out letter by letter. |
+| `<span class="stamp-text">` wrapper | lifts the letters above the ink-void overlay (`z-index`) |
+| A real reference in `.stamp-docket` | a made-up docket number is exactly the authoritative-sounding lie the stamp exists to prevent |
+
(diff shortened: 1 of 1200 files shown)


PipFoweraker:
**Shipping on Pip's ruling (2026-07-31, variant B).** The 1,000 `alignmentforum_*` pages are explicitly *not* blocking this — that gap is now tracked as its own issue rather than living only in this PR body.

### Rebase

`main` moved three times underneath this branch while it waited (#186, #234, #182). Rebased onto current `main`; only `events-sync-summary.json` conflicted, and `sync-events.py` merged cleanly with all three. All 1,194 pages were then **regenerated** from the merged generator rather than trusting the textual merge — and the regenerated output came back **byte-identical** to it, so the merge was already right.

Compatibility with what landed in between, verified on the regenerated tree:
- **#182** (DNT shim + share cards): 1,194 / 1,194 pages still carry exactly one non-deferred consent shim above the deferred tracker and all nine share-card tags, parsed with `html.parser`.
- **#234** (one escaper): `python scripts/test-sync-events.py` passes, 55 assertions.
- **#186**: `og:image` resolves to the new 1200x630 card.
- `public/data/events.json` unchanged.

### The change, as served by the preview

`GET /events/anthropic_exodus_2021.html`:

```html
<span class="stamp stamp--restricted stamp--sm">Not verified in game</span>
	<th>Direction</th>
	<td class="impact-negative">proposed: down</td>
	<td class="impact-positive">proposed: up</td>
```

`GET /css/stamp.css` → `HTTP 200, 8511 bytes, text/css` — the stylesheet the stamp needs is reachable, not a dangling reference.

Coverage, counted rather than asserted:

```
pages carrying the stamp:                        1194
pages still rendering a numeric impact:          1000
  ...of which are NOT alignmentforum_*:             0
```

Exactly the split this PR describes. Nothing outside the E-0 set was missed.

### Two things Pip should know landed here

1. **The known gap is now issue #TBD** (linked below), carrying the finding this PR made: the 1,000 pages come from pdoom-data's `timeline_events/alignment_research/`, which `sync-events.py` has never read; the **schema matches** (`impacts`, `pdoom_impact`, `rarity`, `sources`), so it is cheaper than `TECH_DEBT` E-0 assumes; and the reason it was not done here is that a first regenerate from a months-unread source could move titles, descriptions, categories and sources, not just the impacts block — it wants a diff review, not a unilateral regenerate.

2. **This branch also carries `de94edf0` — "stop calling the game open source (closes #200)"** — which is reader-facing prose on `/` and `/about/`, not event-page work. It is a truth fix ("MIT-licensed" → "source available", plus an FAQ answer that says plainly *"Open source: not yet, and it would be wrong to say otherwise"*), so it belongs under the prime directive. Flagging it because it is a separate change riding in on this PR's ruling, and it will show up in the `snapshot-copy.py --check` advisory as copy drift on two pages — that drift is this, and it is deliberate.

PipFoweraker:
Tracking issue for the known gap is **#248** — the 1,000 `alignmentforum_*` pages, why they are not dead weight, the matching-schema finding, and why this wanted a diff review rather than a unilateral regenerate. Cross-referenced to `docs/TECH_DEBT.md` §E-0 and the Chesterton's-fence note in `CLAUDE.md`.

PR #228, opened 2026-07-31
fix(encoding): three subprocess calls decoding with the locale codec

PipFoweraker:
**The Encoding Safety guard from #198 caught these — the guard working, not breaking.**

Three `subprocess.run()` calls in text mode with no `encoding=`, so the child's output is decoded with the **locale codec**. On a cp1252 console a child writing UTF-8 raises `UnicodeDecodeError` — same family as the trap that has bitten this repo repeatedly.

All three arrived in files that landed alongside or after #198 was written (`sync-keybinds` from #197, the two tests from #185 and #207), so **the guard found real drift within hours of shipping**. That's a good sign about the guard, not a bad one about the code.

`check-encoding-safety.py` now: `OK: 48 of 51 Python modules are encoding-safe (3 waived)`. The three waivers are pre-existing and held by other branches.

Not league-blocking. Fixing it because `main` was red, and a red `main` is the thing we've spent two days removing.

### DIFF
--- a/scripts/render-review-index.py
+++ b/scripts/render-review-index.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python
+"""Render REVIEW_INDEX.md to a clickable local page.
+
+Local review aid only. Writes to public/_review/index.html, which is
+deliberately NOT committed -- see the .gitignore entry this script adds.
+The point is the links: a walkthrough you can click through beats one you
+copy-paste into a viewer.
+
+    python D:\\Local_Code\\pdoom1-website\\scripts\\render-review-index.py
+    -> http://localhost:8080/_review/
+"""
+import html
+import re
+import sys
+from pathlib import Path
+
+for _s in (sys.stdout, sys.stderr):
+    try:
+        _s.reconfigure(encoding="utf-8", errors="replace")
+    except (AttributeError, ValueError):
+        pass
+
+ROOT = Path(__file__).resolve().parents[1]
+SRC = ROOT / "REVIEW_INDEX.md"
+OUT = ROOT / "public" / "_review" / "index.html"
+
+
+def inline(t):
+    """Inline markdown -> HTML. Escape first, then re-introduce markup."""
+    t = html.escape(t, quote=False)
+    t = re.sub(r"`([^`]+)`", r"<code>\1</code>", t)
+    t = re.sub(r"\[([^\]]+)\]\(([^)]+)\)", r'<a href="\2">\1</a>', t)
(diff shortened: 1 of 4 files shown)


PipFoweraker:
Not merged — held, with a reason, after working the merge queue on 2026-08-04.

**The encoding half has already landed.** All three `subprocess.run(..., text=True)` call sites this PR fixes now carry `encoding="utf-8", errors="replace"` on `main` (`scripts/sync/sync-keybinds.py:169`, `scripts/test-snapshot-plausible.py`, `scripts/test_ingest_scores.py`), applied by other merges. `python scripts/check-encoding-safety.py --check` on current `main` reports **0 R3 findings** and exits 0. Rebasing this branch produces content conflicts in two of the three files that are 'the same fix, written twice'.

**What is left is not what the title says.** After discarding the redundant hunks, the only remaining content is `scripts/render-review-index.py` (221 lines) — a local review aid whose sole input is `REVIEW_INDEX.md` at the repo root. **That file exists nowhere**: not on `main`, not on any branch, not in Pip's working tree. Landing it puts a script on `main` that exits on its first read on every possible run, which is exactly the 'a script that can only refuse is indistinguishable from broken' shape CLAUDE.md warns about — and `#227` has since landed `scripts/render-content-review.py`, a second review renderer, so this would also be the second one.

Left open rather than closed so the tool is not thrown away: it is the generator behind the stale `public/_review/index.html` (title `integration/2026-07-29-review`). If `REVIEW_INDEX.md` comes back, retitle this PR and merge just the tool. If it does not, close it.

PipFoweraker:
Closing on Pip's ruling, 2026-08-06.

Verified before closing rather than taking the PR at its word: **all three `subprocess.run(..., text=True)` encoding fixes this PR advertises are already on `main`** — they landed via other PRs while this one waited, and `check-encoding-safety.py` reports **0 R3 findings, exit 0** on main today.

What remained after discounting those was `scripts/render-review-index.py` (221 lines), whose sole input `REVIEW_INDEX.md` exists **nowhere** — not on `main`, not on any branch, not in the working tree. It would exit on its first read on every possible run. It would also be a *second* review renderer beside `scripts/render-content-review.py`, which #227 landed.

So the useful half shipped and the remaining half is a script that cannot run against a file that does not exist. Closing rather than merging.

Recorded because it is the interesting bit: this PR's value was silently transferred to other PRs over four days, and nothing noticed. The branch stayed open advertising work that was already done — a PR describing a world that had moved, which is the same shape as the four stale CLAUDE.md claims and the twenty TECH_DEBT rows corrected in #266 today.

PR #234, opened 2026-08-01
test: cover the three highest-blast-radius untested generators

PipFoweraker:
## What this is

An audit found ~22 scripts that write reader-facing data or published HTML with **no test**. This covers the top three by blast radius, and fixes the live defects the tests exposed.

### The naive cross-reference overstates coverage

`test-orchestrator.py` and `test-integration.py` name five writers between them — `update-version-info.py`, `calculate-game-stats.py`, `health-check.py`, `verify-deployment.py`, `export-leaderboard-bridge.py`. Both only **shell out to each script against the live GitHub API, write into the real `public/` tree, and check the exit code**. Neither asserts a byte of output; a run that fetched garbage and published it passes both. `test-changelog-structure.py` asserts four committed files exist, which says nothing about the generator that wrote them. **No workflow runs any of the three.**

Also found: **`check-published-emails.py` is run by no workflow.** So `redact_pii()` inside `sync-events.py` is the only PII guard that actually executes.

## Ranking (blast radius = pages × silent-failure probability × time-to-notice)

| # | Script | What breaks | Who notices | How long unnoticed |
|---|---|---|---|---|
| **1** | `sync/sync-events.py` | ~1,194 published pages + `events.json`; daily cron **commits**. Owns `redact_pii()`, the only executing PII guard | Nobody. Downstream re-check is unwired | **Indefinitely** |
| **2** | `update-version-info.py` | `version.json` + `status.json` + rewrites `public/index.html`; two 6-hourly crons **commit**. Feeds download buttons, `/game-changelog/`, and `check-platform-claims.py` | A wrong version: Pip, ~1 day. A fabricated `game_stats`: nobody | Days → **indefinitely** |
| **3** | `health-check.py` | `public/data/health-check-results.json`, 6-hourly cron **commits**; publishes maintainer's absolute paths | Nobody reads it | **Indefinitely** (already happened once) |
| 4 | `build-blog-index.py` / `generate-feeds.py` | blog index + RSS/Atom | Subscribers; dead links can't be un-sent | Days |
| 5 | `log-automation-run.py` | `public/monitoring/*`; **fails open** on every write, copies arbitrary `--key value` into published JSON | Nobody | Indefinitely |
| 6 | `generate-sitemap.js` | 2,247 URLs; not committed, generated at deploy | Search Console | Weeks |
| 7 | `stamp-league-epoch.py` | ~45 weekly archive files | `--check` (self-check, not a test) | Weeks |
| 8 | `sync/merge-alignment-events.py` | writes **upstream** `all_events.json`, which fans out to ~2,194 pages; **no PII redaction** on that path | Nobody | Indefinitely |
| 9 | `update-game-status.js` | `status.json`; **no safety property at all**, invents a website version bump each run | Nobody | Indefinitely |
| 10 | `validate_data.py`, `snapshot-copy.py`, `check-stale-facts.py` | the guards themselves — untested, so "green" is equally consistent with "never fires" | — | — |

Remaining untested writers, lower blast radius: `calculate-game-stats.py`, `generate-metabolism.py`, `sync/sync-keybinds.py`, `sync/sync-game-icons.py`, `prepare-weekly-deployment.py`, `prepare-syndication.py`, `post-syndication.py`, `generate_roadmap_from_milestones.js`, `generate_pdoom1_issues_md.js`, `update-version-info.js`, `check-board-liveness.py`, `board_liveness_summary.py`, `game-integration.py`, `verify-deployment.py`, `extract_analytics.py`.

## Live defects the tests exposed (all now fixed)

**In the shipped events corpus, on production:**
- `arxiv_73643a60bb86bf2f` — description contains `<<number to be assigned>>`, parsed as a tag start in `<p class="description">` and inside `<meta name="description" content="...">`
- `arxiv_aa8c44de8cf70353` — a `"` inside the first 155 characters **terminated the meta content attribute early** and turned the rest of the sentence into bogus tag attributes

The generator escaped nothing. arXiv accepts uploads from anyone, so "the data is first-party" is not true of the *contents* of these fields.

**In `update-version-info.py`:**
- `.AppImage` contains `.app` → a Linux-only release published `platforms.macos: true`
- `x86_64` is an *architecture* → `PDoom-<version>-macos-x86_64.dmg` published `platforms.linux: true`
- Both fed `check-platform-claims.py` — the guard whose entire job is stopping a false platform claim — a false positive
- `game_stats` was hardcoded to `{23, 7, 10000}` every run, overwriting the honest `null` + `"not yet measured"` block `calculate-game-stats.py` writes to the same file with the comment *"DO NOT restore a literal here 'temporarily'. That is exactly how these two survived for months."* **This script was the thing restoring them.**
- `tag_name` was passed as `re.sub()`'s **replacement template**, so `\1` / `\g<0>` in a tag expanded and a lone backslash raised `re.error` mid-rewrite of the homepage

**In `health-check.py`:**
- Four `except Exception as e:` handlers interpolated the raw exception string. `str(OSError)` embeds the absolute path it failed on — re-publishing the exact class of string `scrub()` exists to remove, through the guard, on a cron, to a public URL
- `_ABS_PATH` was **inert against its own founding case**: `[^\s'"]+` stops at a space, so `C:\Users\<name>\Documents\A Local Code\...` — the path CLAUDE.md quotes as the leak — scrubbed to `A Local Code\pdoom1-website\public\data\version.json`

## Evidence each test can fail

Every guard was broken on purpose and observed failing, then restored.

| Break | Result |
|---|---|
| Remove `escape_event_for_html()` from the generator | **9 failures**, incl. both named production regressions and `evt_html.html :: raw script tag` in the end-to-end scan |
| Rebuild `redact_pii()` as a four-field list (the plausible "optimisation") | **5 failures**, incl. `THE BIG ONE: an unknown field added upstream tomorrow is redacted today` and `events.json :: address` |
| Restore the hardcoded `game_stats` literals | **5 failures**, incl. the source scan: `found [('baseline_doom_percent','23'),('frontier_labs_count','7'),('strategic_possibilities','10000')]` |
| Revert `re.sub` to a replacement template | **8 failures**: `invalid group reference 1`, `bad escape (end of pattern)`, and two `cannot open an element on the front page` |
| Remove chokepoint scrubbing from `log_result()` | **21 failures**, incl. a real leak in the end-to-end run: `Permission denied: 'C:\Users\gday\AppData\Local\Temp\...'` |

Before the `_ABS_PATH` fix, the space-containing-path assertion failed on its own.

## Rules, not enumerations

- **sync-events**: renders the same event twice — benign text in every string slot, then a hostile payload in every string slot — and requires an **identical tag-and-attribute skeleton** (`HTMLParser`). No field is named. A new interpolation that forgets to escape fails on the day it is written. Also asserts, as a rule, that **every attribute in the template is double-quoted**, which is what makes leaving apostrophes unescaped sound.
- **redact_pii / escape_event_for_html**: both fed a field that does not exist in today's schema.
- **health-check**: scans **every string anywhere in the published JSON** against `_ABS_PATH`, not four named messages.
- **update-version-info**: a source scan rejects any numeric literal assigned in `update_version_data()`; the platform table asserts "result keys == declared platform keys" rather than listing three.

## No moving values compared to literals

Fixture versions are synthetic (`v9.9.9`, `v1.2.3`) and every rule reads the value **out of the fixture it built**. Nothing pins a date, a release, a file count or a success percentage. The three files join `check-stale-facts.py`'s existing `SKIP_FILES` for the same reason `board-liveness.json` is there — their version strings are inputs, not claims. **Verified: 214 findings before and after.**

## Isolation

Every case builds inputs in `tempfile.mkdtemp()` and redirects the module's path constants there. No committed fixture is mutated, no test touches `public/`, and **nothing issues an HTTP request** — `fetch_github_data` is stubbed, and the events sync is handed a fake `pdoom-data` tree.

## Review note: diff size on the next sync

The escaping fix means the next events sync rewrites all 1,194 pages, but the diff is **one line each** — a bare `&` in the `mailto:` href becoming `&amp;` (correct HTML) — plus real escaping on the handful of events with markup in their text. `esc()` deliberately does **not** escape apostrophes, precisely so that diff stays reviewable. Verified: all 1,194 real events regenerate with zero errors.

## Test results

All three pass; 171 assertions total. Every other test in CLAUDE.md's suite is unchanged from the base branch — the seven that fail (`validate_data`, `check-stale-facts`, `check-published-emails`, `snapshot-copy --check`, `generate-metabolism --check`, `test-header-consistency`, `check-encoding-safety`) fail **identically on the base commit**; confirmed by stashing. `generate-metabolism.py` still builds cleanly, so no citation needle was broken by these edits.

## Not done

- Tests for ranks 4–10. `merge-alignment-events.py` is the one I would take next: it writes the **upstream** `all_events.json` that regenerates ~2,194 pages, applies **no PII redaction** on that path, and is run by no workflow — so the redaction fixed here can be undone from the other side.
- `check-published-emails.py` is still wired to no workflow. Wiring it to `content-honesty.yml` is a one-line change and would give the PII guard a second, independent check.
- `sync-events.py` still carries an encoding-safety waiver (`W1`, no stdout preamble — it uses a `win32` `TextIOWrapper` instead, which works but the checker doesn't recognise).
- `update-version-info.js` is an orphan duplicate writer of `version.json` that does **not** write `latest_release.platforms`. Left alone; deleting it is a separate call.

### DIFF
--- a/.github/workflows/content-honesty.yml
+++ b/.github/workflows/content-honesty.yml
@@ -42,6 +42,13 @@ on:
       - 'scripts/test-platform-claims.py'
       - 'scripts/check-published-emails.py'
       - 'scripts/check-stale-facts.py'
+      # The three committing generators and the tests that force their refusals.
+      - 'scripts/sync/sync-events.py'
+      - 'scripts/test-sync-events.py'
+      - 'scripts/update-version-info.py'
+      - 'scripts/test-update-version-info.py'
+      - 'scripts/health-check.py'
+      - 'scripts/test-health-check.py'
       - 'scripts/snapshot-copy.py'
       - 'scripts/snapshot-plausible.py'
       - 'scripts/test-snapshot-plausible.py'
@@ -76,6 +83,13 @@ on:
       - 'scripts/test-platform-claims.py'
       - 'scripts/check-published-emails.py'
       - 'scripts/check-stale-facts.py'
+      # The three committing generators and the tests that force their refusals.
+      - 'scripts/sync/sync-events.py'
+      - 'scripts/test-sync-events.py'
+      - 'scripts/update-version-info.py'
+      - 'scripts/test-update-version-info.py'
+      - 'scripts/health-check.py'
+      - 'scripts/test-health-check.py'
       - 'scripts/snapshot-copy.py'
       - 'scripts/snapshot-plausible.py'
       - 'scripts/test-snapshot-plausible.py'
@@ -138,6 +152,21 @@ jobs:
       # Third parties' addresses scraped out of paper PDFs must never be served.
       - name: No third party's email address is published
         run: python scripts/check-published-emails.py
+
+      # The three generators that publish reader-facing data on committing crons.
+      # check-published-emails.py above walks what is ALREADY COMMITTED, so it can only
+      # report a leak that has already shipped; these force the failure at the generator,
+      # which is the only place it can still be prevented. All stdlib, all offline --
+      # each stubs its inputs in a temp dir and asserts the refusal, so none of them
+      # needs network, a pdoom1 checkout, or the `requests` this job does not install.
+      - name: sync-events redacts PII and no event text can reshape a page
+        run: python scripts/test-sync-events.py
+
+      - name: update-version-info refuses to guess, and derives platforms from assets
+        run: python scripts/test-update-version-info.py
+
+      - name: health-check publishes no absolute path
+        run: python scripts/test-health-check.py
 
       # --ci, not --check: drift and freshness are only fixable by re-running the sync
       # against a local pdoom1 checkout, which no runner has, so they print as WARN.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -210,6 +210,22 @@ Pip's stated top priority. Practically:
   so a new upstream field cannot leak). `scripts/check-published-emails.py`
   re-verifies and imports the generator's one regex rather than copying it.
   The addresses are still in pdoom-data — scrubbing here does not fix the source.
+  - **`check-published-emails.py` checks the OUTPUT, not the generator.** It walks
+    what is already committed under `public/`, so it can only tell you a leak has
+    already shipped. `redact_pii()` is the thing that stops one. That is why
+    `scripts/test-sync-events.py` forces the leak rather than watching the guard
+    pass. (An earlier draft of this bullet said the guard "is run by no workflow,
+    verified 2026-08-01" — that was true when written and is now false:
+    `content-honesty.yml:121` runs it as a BLOCKING step. Grep before believing
+    a CI claim here, in either direction.)
+- **The event page template escapes nothing by itself** — it is one 500-line
+  f-string, so the data decides where the markup ends, and arXiv descriptions are
+  raw PDF text uploaded by anyone. `escape_event_for_html()` walks the whole
+  record (same shape as `redact_pii()`) and runs once at the top of
+  `generate_event_detail_page()`. `esc()` escapes `& < > "` but **not** the
+  apostrophe, because every attribute in the template is double-quoted — a rule
+  `test-sync-events.py` asserts, so writing `style='…'` in the template fails the
+  test rather than silently reopening the hole.
 - **Events were not the only mirror.** `public/data/issues-cache.json` is a verbatim
   copy of pdoom1's issue bodies, rendered by `/issues/`, and it was written raw. On
   2026-08-02 a `Co-Authored-By` trailer inside an issue body put a live address on the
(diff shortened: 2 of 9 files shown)


PipFoweraker:
Rebased onto current `main` (was 9 behind, 3 ahead, with a merge commit in the middle). Now two linear commits on `main`, force-pushed with `--force-with-lease`.

**Conflict resolution** — both conflicts were in `CLAUDE.md`, and both were additive on each side, so the resolution is the union in both cases:

- *Local test suite list*: `main` had restructured the list with `[workflow]` brackets (the `ci/wire-the-guards` audit). This branch still carried the pre-audit copy, unfenced, plus three genuinely new lines. Took `main`'s structure and re-added only the three new entries, in the bracket convention.
- *`check-platform-claims.py` findings*: kept both bullets — the asset-filename false positives (this branch) and the vacuous early return (`main`). They describe different bugs.

The third commit, `0dfcf084` "resolve CLAUDE.md conflict markers pushed by mistake", was **skipped**: its entire content was removing markers left by the merge commit that the rebase dropped, taking the same union resolution applied above. Nothing was lost.

**One commit added**: `chore(metabolism): regenerate`. #232 renamed `post-issue-to-forum.yml` to `.disabled` without rebuilding `/metabolism/`, so the source count moved 28 -> 27 and `metabolism-map.yml` had been failing on every PR since. That check is green now.

**Blocked on #239.** The one remaining red is `check-stale-facts.py --min-severity HIGH`, and it is not this branch's: it reproduces on a pristine `main` checkout. The single HIGH is `public/data/issues-cache.json:169` — a bot-mirrored pdoom1 issue body quoting a "v0.13 playtest" note. #239 clears it by adding `issues-cache.json` to `SKIP_FILES`.

Note for whoever merges second: **#239 and this PR both add to `SKIP_FILES` immediately after `board-probe-targets.json`**, so they will conflict textually. The resolution is the union — the two entries are unrelated.

Everything this PR owns is green: all three new test files pass locally and in CI, and 8 of 8 non-skipped checks pass apart from the shared gate above.

PR #236, opened 2026-08-01
fix(security): sweep the XSS sink class off the remaining 14 pages, and consolidate five escapers into one

PipFoweraker:
`/leaderboard/` was hardened over 2026-07-30/31 (#202, #208) after a live stored-XSS hole. The rest of the site was not. This is that sweep, plus the consolidation and the test that makes it stick.

## Reported findings: what I CONFIRMED

Each was treated as a lead and checked against the code before acting.

| Reported | Verdict | Evidence |
|---|---|---|
| `dashboard/` — `market.question` into `innerHTML`, `market.url` into `window.open` with no scheme check | **CONFIRMED, worst one** | `public/dashboard/index.html:1525` and `:1523`. Third-party text on a top-level nav page. |
| `issues/` — escapes one of five GitHub fields; `label.color` lands in a `style=` attribute | **CONFIRMED** | `public/issues/index.html:535` (`style="background: #${label.color}"` plus `${label.name}`), `:550` (`href="${issue.html_url}"`), `:558` (`${issue.number}`). Only `title` was escaped, at `:551`. |
| `league/index.html` — no escaper defined at all | **CONFIRMED** | `player_name`, `game_mode`, `level_reached` raw at `:927`, `:929`, `:934`. |
| `players/index.html` — no escaper | **CONFIRMED** (narrower) | One external sink: `${week.week}` at `:920`. The rest of the page is `textContent`. |
| `monitoring/index.html` — no escaper | **CONFIRMED** | `check.test`, `jobName`, `weekInfo.*`, `league.seed`, `c.name`, `c.detail`, `h.overall_status` all raw. |
| `dev-notes/` — no escaping in the markdown renderer | **CONFIRMED** | `public/dev-notes/index.html:53-62` put raw `/docs/DEV_NOTES.md` into `innerHTML` and handed `$2` straight to `href=""`. |
| `blog/post.html` + `docs/roadmap/` — holed markdown renderers | **CONFIRMED** | Both escapers covered `& < >` and not quotes, while their link rules accept `[^)\s]+` as a URL — a `"` included. `[x](a"onmouseover=alert(1))` broke out of the `href`. |
| Three differently-named escapers, at least two not covering quotes | **CONFIRMED, and undercounted** | There were **five**: `escapeHtml` (blog/post), `esc` (roadmap), `esc` (game-changelog), `escapeHtml` via `textContent` (issues), `escapeHtml` (league/archive). Four fed attribute contexts; only league/archive's covered quotes. |

## What did NOT survive checking

**`league/archive.html` "has no escaper defined at all" — false.** It had the only *complete* escaper on the site (`:731`, all five entities, quotes included). The real defect is different and arguably worse: it was applied to **9** interpolations and skipped on **11**, including `winner.player_name` in the click-to-open toast (`:1055`, `showToast` → `innerHTML`) and every field of the card grid. A partially-applied escaper defeats a reviewer's spot check — the file *reads* as protected.

That lead being wrong changed the fix: the page did not need an escaper added, it needed the existing calls audited.

## Found while sweeping, not reported

**`.toFixed()` / `.toLocaleString()` on raw API fields is a denial of service, and `/leaderboard/` still had eight of them.** `entry.final_doom` is passed straight to `.toFixed(1)` at `leaderboard/index.html:1129` and `:1354`; `(entry.final_reputation || 0).toFixed(1)` at `:1167` and `:1423`. `|| 0` does not help — a non-empty string is truthy, and `String.prototype` has no `toFixed`. One `TypeError` inside a render loop takes down the **whole** board, not one row. A single `POST {"score":"x"}` blanks `/leaderboard/`, `/league/` and `/league/archive.html`, with no injection at all. Fixed with `toNumber()` at every such site.

## Part 2 — consolidation

`public/assets/js/escape.js` is now the **only** escaper. Five became one.

- `escapeHTML` — element text **and quoted attributes** (covers `"` and `'`, which three of the five did not).
- `safeUrl` — a quoted `href`/`src` in an HTML string. Scheme-checked *and* escaped.
- `safeUrlRaw` — a JS sink (`window.open`). Scheme-checked only; `safeUrl` would turn the query `&` into `&amp;`.
- `isSafeUrl` — the bare predicate, for a URL already escaped in place (the markdown renderers).
- `toNumber` — the availability fix above.

Escaping alone can never make a URL safe: `javascript:alert(1)` contains no HTML metacharacter. The scheme test runs against a control-character-stripped probe, because browsers strip those *inside* a scheme (`java<TAB>script:` is `javascript:` to the parser).

Loaded by a **plain blocking** `<script src>` in each page's head. If it fails to load, `escapeHTML` is undefined, the template throws, and the page renders nothing — **fail closed**. The test asserts it is not `defer`/`async`.

Two contexts deliberately **not** solved by escaping: `label.color` goes inside `style="background:#…"`, which is CSS — a `}` or `;` needs no HTML metacharacter — so it is validated as six hex digits instead. `getContrastColor()` also used to throw on a label with no colour and take the grid down.

## Part 3 — the test enforces the RULE

`scripts/test-escaping.js`. The first version of `test-board-escaping.js` named six fields and passed while thirteen were unprotected, so nothing here is a field list.

1. **One escaper.** Any page defining its own fails. Also catches hand-rolled `.replace(/</g,'&lt;')` — which `/website-changelog/` had, applied to two of five fields.
2. **Every interpolation of external data is escaped — by ROOT, not by field.** Each page declares the identifiers holding parsed `fetch` results (`entry`, `market`, `issue`, …). A **new field on an existing root is covered the day it is added**, with no test edit.
3. **The root list cannot rot silently.** Rule 2 cannot cover a new *data source*, because that arrives with a new variable name. So every `fetch()` URL on a guarded page must be declared — add a fetch and the test fails until someone says where its result lands. That is the loop a field-list check structurally cannot close.
4. **Markdown renderers are run, not grepped.** Each is extracted from its own page, given the shared escaper the way the browser does, and fed 12 hostile inputs.
5. **No numeric method on a raw API field.**

The predicate is **structural**: after removing the markup a producer is entitled to emit, no `<`, `>` or bare quote may survive — not a keyword blocklist. Escaping does not delete the letters that spell `onerror`, it makes them inert, so a name search answers a different question. Dangerous schemes are checked **only inside extracted `href`/`src` values**, because `javascript:alert(1)` sitting in the body as escaped *text* is the correct outcome of a rejected link.

**Forced-failure evidence** (CLAUDE.md: "a guard seen only in its passing state has not been shown to work"). All three verified, then reverted:

```
FAIL  league/index.html DEFINES ITS OWN ESCAPER: escapeThing
FAIL  league/index.html: 1 UNESCAPED -> ${entry.brand_new_field_2027}
FAIL  league/index.html: UNDECLARED fetch -> /data/some-new-source.json
```

The test also carries a permanent forced failure: it constructs the `& < >`-only escaper the three pages used to have and asserts it is **rejected** on an attribute payload while **passing** on an element payload — which is exactly why nobody noticed for months.

`.github/workflows/escaping.yml` runs both escaping tests in CI. **Neither had ever run there**, and the board guard was additionally inert on Windows via CRLF — so the one check protecting an unauthenticated API was silent in both places at once.

## Verification

- `test-escaping.js`, `test-board-escaping.js`, `test-blog-render.js`, `test-roadmap-render.js`, `test-changelog-render.js` — all pass. The last four were updated to load the shared escaper, mirroring the page's real dependency graph.
- Every inline `<script>` on all 15 touched pages re-parsed with `vm.Script`.
- **Zero regressions.** Seven suite checks were already red (`validate_data`, `check-stale-facts`, `check-published-emails`, `snapshot-copy --check`, `generate-metabolism --check`, `test-header-consistency`, `check-encoding-safety`). I ran all of them against a pristine `git archive HEAD` extraction: **byte-identical output**. None touch a file in this PR. `snapshot-copy --check` being identical also confirms no reader-facing prose changed.

## Judged too risky to change without Pip

- **`/dashboard/` reads `/data/game-changes.json` (`:1465`), which CLAUDE.md records as "read by nothing" and which carries a `_deprecated` note.** One of those two is wrong. I escaped the fields rather than resolving it — whether that dev-log panel should exist is a product call.
- **Migrating `/leaderboard/` to the shared file is the highest-risk edit here.** It adds a load-order dependency to the site's most security-sensitive page. Done because leaving it inline means two implementations that can drift — the exact failure the one-escaper rule exists to stop — but it is the change to eyeball on the deploy preview first.
- **`safeUrl` rejects protocol-relative `//host/x`.** Nothing on the site uses one, but it is stricter than strictly necessary.
- **`/league/`, `/players/` and `/league/archive.html` are retired-and-hidden** (Pip, 2026-07-28). Fixed rather than deleted — `rsync --delete` makes deletion a production removal — but if they are going, that is cheaper than maintaining them.
- **Not swept:** `public/events/*.html` (~2,194 generated pages) and `public/stats/competition.html`. The event pages render no fetched data; competition.html renders only numbers into `textContent`, though it does share the `.toFixed()` pattern on `statusData?.avg_score`. Worth a follow-up, not silently bundled here.


PipFoweraker:
### Merged `main`, and one thing that only showed up after doing so

#233 (`ci: wire the 10 unwired pre-PR guards`) merged while this was open and landed a **new** test, `scripts/test-board-honesty.js`, which also pulls `function escapeHTML(s)` out of `public/leaderboard/index.html` by regex. This branch moved that function into the shared file, so it died at extraction — `could not extract escapeHTML() from the page`, exit 1 **before a single assertion ran**. That is the failure mode `test-board-escaping.js` already had once via CRLF: not a red test, an absent one that looks like a test.

Fixed the same way as the other four: it loads the shared module and takes it as a function parameter, which is the node-side equivalent of the page's blocking `<script src>`.

Two more things that fell out of the merge, both worth stating because they change what this PR does:

- **`escaping.yml` no longer duplicates anything.** #233 gave `test-board-escaping.js` a home in `board-liveness.yml`, `test-changelog-render.js` one in `content-honesty.yml`, and `test-blog-render.js` was already in `generate-feeds.yml`. Running any of them twice doubles the noise and halves the trust in either signal, so this workflow now runs only `test-escaping.js` (site-wide) and `test-roadmap-render.js` — which, after #233's audit, was still wired to **no workflow at all**.
- **`public/assets/js/escape.js` is now in the path filter of `board-liveness.yml` and `generate-feeds.yml` too.** Every renderer on the site depends on that one file now. A change to it that only re-ran *this* workflow would be a hole of exactly the kind #233 was cleaning up.
- `/metabolism/` regenerated: it derives from `.github/workflows/`, and was stale for two reasons — this branch's new workflow, and main's own rename of `post-issue-to-forum.yml` to `.disabled`.

### CI: green except one pre-existing red

`Board guards (offline)`, `One escaper, every external field escaped`, `check` (metabolism), `feeds`, `health-check`, `check-version`, `Static sweep (AST)`, `Real scripts under a cp1252 stdout`, the Netlify preview — all pass.

**`BLOCKING -- reader-facing claims` fails, and it is not this PR.** The single HIGH finding is `public/data/issues-cache.json:95` — a bot-refreshed cache quoting Pip saying *"v0.13 playtest 2026-07-24"* inside an issue body. I extracted pristine `origin/main` with `git archive` and ran the gate against it: **215 findings, 1 at or above the HIGH gate — identical**. `main` is red on this right now.

I have deliberately not fixed it here. It is a generated cache file, so an edit would be reverted by the next `pull-pdoom1-issues` run; the real fix is either in that generator's severity treatment or in `check-stale-facts.py` learning that a quoted playtest note is history, not a fallback. Bundling that into an XSS PR would hide it.

PR #274, opened 2026-08-06
chore(stats): retire public/stats/ -- DreamHost 401s it, and it was game-stats twice

PipFoweraker:
Retires `public/stats/`. Approved by Pip 2026-08-06; this is the product call TECH_DEBT **A8** had been holding open.

## The decisive fact: nobody has ever been able to load it

DreamHost's shared hosting owns `/stats/`. It is the panel's own web-stats area, configured above the document root, and it shadows ours entirely.

Measured against production 2026-08-06:

| URL | result |
|---|---|
| `https://pdoom1.com/stats/` | **401** — `WWW-Authenticate: Basic realm="Statistics Area"` |
| `https://pdoom1.com/stats/competition.html` | **401** — same realm |
| `https://pdoom1.com/game-stats/` | **200** |

`public/stats/.htaccess` held an `AuthType None` / `Satisfy Any` override written by someone earlier for exactly this purpose. It deployed, and the 401 above is what it achieved. The host wins at a level a directory `.htaccess` cannot reach.

## It was also the same page twice

`diff public/stats/index.html public/game-stats/index.html` against **current `main`** (i.e. after #196 landed today) — **14 differing lines of ~310**:

- the `<link rel="canonical">` URL
- two stylesheet links `/game-stats/` gained in #196 (`game-integration.css`, `ui-enhancements.css`)
- whitespace and comment wording around the nav placeholder and the `navigation.js` tag

No content difference. The cost was real and paid this week: the `23%` self-contradiction and the frontier-labs count were each **two edits for one visible page**.

## `competition.html` was NOT a duplicate — read this bit

The brief described `/stats/` as a duplicate. That is true of `index.html` only. `stats/competition.html` was a **distinct** Competition Statistics page with no equivalent at `/game-stats/`, so this PR **retires** it rather than replacing it.

Retiring it is still within the approved decision — TECH_DEBT A8 and `robots.txt` both name *both* files as the open retire-or-keep question — and it was equally unreachable behind the same 401. But it is the one item worth a second look before merge.

Recovery, if it is ever wanted:

```
git show <this-PR's-merge-commit>^:public/stats/competition.html
```

Reviving it means first choosing a path DreamHost does not shadow.

## The 301 — unverifiable before merge, and expected to lose

`public/.htaccess` now attempts `RewriteRule ^stats(/|/index\.html)?$ /game-stats/ [R=301,L]`, following the `/changelog/` precedent. The `^` anchor stops it looping and deliberately does not match `competition.html` — sending a URL named "Competition Stats" to the game-stats page would promise one thing and deliver another.

**Netlify ignores `.htaccess`, so the deploy preview cannot exercise this rule.** And the prior should be that it does not fire: a docroot `RewriteRule` is a *weaker* lever than the directory-level auth override that already lost. Shipped anyway because the downside is nil — the path already fails closed, so a rule that never runs leaves things exactly as they are, and if the panel's stats area is ever switched off the URL lands somewhere honest.

**Verify on production after merge:**

```
curl -sI https://pdoom1.com/stats/
#   301 + Location: /game-stats/  -> the rule took
#   401 realm "Statistics Area"   -> DreamHost still owns it (expected; no follow-up needed)
```

Either outcome is acceptable. The caveat and this command are recorded in `docs/TECH_DEBT.md` under "A8 caveat" so it is not re-litigated.

## Every `/stats/` reference found and fixed

Swept with `grep -rn 'href="/stats' public/` and `grep -rn 'public/stats\|/stats/' scripts/ docs/ .github/ public/`.

| file | was | now |
|---|---|---|
| `public/robots.txt` | `Disallow: /stats/` + a 6-line comment recording that both pages carried no `noindex` | line dropped; comment replaced with why it went. The Disallow was protecting nothing, and a Disallow for a path with no files behind it is a claim about content that no longer exists. The noindex gap could never have been closed by adding a tag — a crawler that gets a 401 never reads one. |
| `public/league/index.html` :589-590 | "Stats ▾" dropdown over `/stats/` + `/stats/competition.html`, both dead | collapsed to a plain `/game-stats/` link — a dropdown holding one item is a link |
| `public/league/index.html` :762 | "Competition Stats" CTA button → `/stats/competition.html` | **removed**, not repointed — see honesty note above |
| `public/league/archive.html` :582 | nav → `/stats/` | → `/game-stats/` |
| `public/players/index.html` :494 | nav → `/stats/` | → `/game-stats/` |
| `docs/TECH_DEBT.md` A8 | `PARTLY DONE` → remainder `NEEDS-PIP` | **DONE**, with the 401 evidence and the post-merge caveat |
| `docs/ORPHANED_PAGES_TODO.md` | "One human path in survives and is intentionally unbroken: `stats/competition.html` links to `/league/`" | struck — the league trio now has **zero** inbound paths from anywhere on the site |
| `docs/LEAGUE_EPOCH_ANOMALY.md` correction 4 | "these two carry no robots meta tag at all" | marked MOOT with the reason |
| `docs/SURFACE_REGISTER.md` §4.2 | "hardcoded in three places … homepage, `/stats/` and `/game-stats/`" | noted that two of the three were the same page twice; this is the mechanical reason the double-edit stops |
| `docs/weekly-league-future-enhancements.md` | tasks targeting `/public/stats/competition.html` | stale banner at the top; the file is a record of intent, not a plan |
| `scripts/generate-sitemap.js` :260 | docstring example `public/stats/x.html` | uses a path that exists |

Checked and needing **no** change:

- `scripts/check-platform-claims.py` `REACHABLE` — names `public/game-stats/index.html`, never `/stats/`. Confirmed by `test-platform-claims.py` §9, which asserts every `REACHABLE` path is on disk.
- `public/sitemap.xml` — `grep -c 'pdoom1.com/stats' → 0`. It never listed the page (the generator parses `robots.txt`, and `/stats/` was disallowed). It is regenerated by all four deploy workflows anyway.
- `public/data/website-changes.json:38` — "Relocated stats page to /game-stats/ to avoid authentication conflicts". Correct history, and the origin story for this PR: the page *was* relocated, the old copy was just never deleted.
- 2,197 generated `public/events/*.html` — zero hits.

## Guards

All exit 0 on this branch:

```
python scripts/generate-metabolism.py --check    OK: /metabolism/ is in step with its sources
python scripts/check-deploy-excludes.py          PASS (2251 deployed files scanned)
python scripts/test-platform-claims.py           OK (10 sections, incl. the forced failure)
python scripts/check-platform-claims.py          OK
python scripts/check-encoding-safety.py          OK: 57/60 (3 pre-existing waivers, untouched)
```

No test was weakened or deleted.

## Note on the deploy

`rsync --delete` makes this a real production removal of three files. That is the intent. Nothing that a visitor can currently reach is being removed, because the 401 means there is nothing there they can reach.


PipFoweraker:
**Heads-up on the one red check.** `BLOCKING -- reader-facing claims` fails on this PR for a reason unrelated to the diff:

```
FAIL: public/data/analytics/history/index.json is out of step with the snapshots.
      Run: python scripts/generate-analytics-index.py
```

That reproduces on **current `main` with no changes at all**. Sequence: `55779e48` committed the 2026-08-06 daily snapshot, then #272 merged and added `generate-analytics-index.py` plus the `--check` gate — so the newest snapshot was written by a run that predated the rebuild step, and the index was one day short the moment the gate went live.

The fix is carried in **#276** (`bc7b486d`). Either merge #276 first and this goes green on a rebase, or wait for `snapshot-analytics.yml`'s 11:30 run, which rebuilds and commits the index itself.

Nothing in this PR touches `public/data/analytics/`.

PR #289, opened 2026-08-07
ci(#285): let the release event drive version.json, not the clock -- proposal from the pdoom1 seat

### LINKED ISSUE #285: Release published between 6-hourly runs leaves pdoom1.com advertising the previous version -- up to 6h, green the whole time
**Found the hard way on the `v0.14.0` cut, 2026-08-07/08.** Pip had already published links to `pdoom1.com` while the site served the previous version.

## What happened

| Time (UTC) | Event |
|---|---|
| 12:21 | `auto-update-data.yml` ran on schedule. Latest release was `v0.13.2`. **Correct at the time.** |
| 12:52 | `v0.14.0` published on `pdoom1` |
| 12:53–12:54 | `sync-game-version` ran and committed (to `data/`, see #NEXT) |
| 13:5x | Deploy ran green. **Site still served `v0.13.2`.** |
| 14:01 | `auto-update-data.yml` **dispatched by hand** → wrote `v0.14.0` |
| 14:0x | Deploy → site correct |

**The schedule is `cron: '0 */6 * * *'`.** A release published one minute after a run leaves the site advertising the previous version **for up to six hours**, with correct-looking green ticks the whole time.

## Why this is worse than a stale cache

**It is not stale — it is confidently wrong.** `public/data/version.json` carried a complete, well-formed, internally consistent record of `v0.13.2` including its full release notes and a `html_url` pointing at `releases/tag/v0.13.2`. Anything rendering it produced a correct-looking page advertising the wrong build, and any visitor following that link downloaded the old game.

Links to `/releases/latest` and the unversioned aliases (`PDoom-Windows.zip` etc.) were correct throughout. **The exposure is specifically the version-pinned surfaces.**

## The fix, and why the obvious one is insufficient

**Insufficient:** shortening the cron. It narrows the window; it does not close it, and it burns runs.

**Suggested:** make the release *event* drive the update rather than the clock.

- `pdoom1` already runs `sync-game-version.yml` at release time. Have it **also** refresh `public/data/version.json`, or dispatch this workflow.
- Or add `repository_dispatch` here and have `pdoom1` fire it on publish.
- **Keep the 6-hourly cron as the backstop**, not as the mechanism.

## Closing observation (write-once, per `coordination#5f`)

This issue may be closed when the following holds **within 10 minutes of a release publishing**, without a human dispatching anything:

```bash
curl -s "https://pdoom1.com/data/version.json?cb=$(date +%s)" | jq -r .latest_release.version
# must equal:
gh release view --repo PipFoweraker/pdoom1 --json tagName --jq .tagName
```

**Note the URL is the live site, not the repo file.** Tonight proved those are different assertions: the repo held `v0.14.0` while `pdoom1.com` served `v0.13.2`.

PipFoweraker:
Proposal from the **pdoom1 seat**, implementing your issue #285 rather than redesigning it. **Yours to accept, amend or reject** -- I have not pushed to `main`, have not merged, and have not closed #285.

#285 offered two shapes and I took the second verbatim:

> - Or add `repository_dispatch` here and have `pdoom1` fire it on publish.
> - **Keep the 6-hourly cron as the backstop**, not as the mechanism.

## What this changes

**`auto-update-data.yml`** gains `repository_dispatch: types: [game_version_sync]`. The cron stays exactly as it is, re-labelled in a comment as the backstop.

**`auto-deploy-on-push.yml`** gains `"Auto-Update Data"` in its `workflow_run` list, plus a `deploy-production` concurrency group.

## What I verified before writing it

| Claim | Command | Result |
|---|---|---|
| Nothing here listens for `game_version_sync` | `git grep -c game_version_sync` in this repo | exit 1, **zero hits**. Five other dispatch types are wired (`sync-tokens`, `pdoom1_release`, `pdoom1_docs_update`, `bug-report`, `events-updated`, `update-game-status`) |
| ...and never has | `gh api "repos/PipFoweraker/pdoom1-website/actions/runs?event=repository_dispatch"` | **empty list.** No `repository_dispatch` run has ever executed on this repository |
| pdoom1 really is sending it | `sync-game-version.yml` "Trigger Website Rebuild" step | POSTs `{"event_type": "game_version_sync", ...}` to `/repos/PipFoweraker/pdoom1-website/dispatches` |
| ...and sees success regardless | GitHub dispatch API | returns 204 whether or not anything is subscribed. **The sending side cannot tell.** That is why this went unnoticed rather than failing loudly |
| The bot push does not deploy | run history, v0.14.0 cut | `version.json` committed 14:01Z; the deploy that carried it to production was `workflow_run` off Board liveness, **not** `push` |
| Deploys can overlap | `gh run list --workflow=auto-deploy-on-push.yml` | two runs on different SHAs at **2026-08-07 07:35:17Z and 07:35:20Z**, `rsync --delete` phases 10s apart. No deploy workflow here declares `concurrency:` |

## The ordering hazard, in the comment and repeated here

`scripts/update-version-info.py` derives `latest_release.platforms` from the release's actual **asset filenames**, and the download page offers exactly what `version.json` says. A dispatch that arrives **mid-upload** publishes a version with missing or wrong platforms. Closing the six-hour window this way makes that race reachable for the first time.

The fix belongs on the **sending** side (fire after assets are attached), and that half is mine. If you would rather this workflow defend itself -- e.g. refuse to write when the release has zero assets -- say so and I will not argue; it is your surface and a receiver-side guard is the more robust of the two.

## What I did not touch, deliberately

**Repo-root `data/current-game-version.json`.** pdoom1's `sync-game-version.yml` clones this repo and writes that file, plus `data/versions/*.json` and `content/game-versions/`. `git grep current-game-version` finds **zero consumers** here, and `data/` is outside the rsync source, so **it cannot change what a visitor sees**. It is a producer with no reader, and it looks like it is working -- the file is present and currently correct at `0.14.0`.

That is the deeper defect and it is your surface. Naming it here is right; fixing it unilaterally in your repo is not. If you want it gone, the pdoom1 side that writes it is mine to remove and I will do that on your word.

## The concurrency group is deliberately incomplete

It only serialises `auto-deploy-on-push` against itself -- which is the case actually measured. `deploy-dreamhost.yml`, `version-aware-deploy.yml` and `weekly-deployment.yml` need the same group string to serialise against each other. I did not make those edits because I do not know which of them you consider live.

## How to prove it works

#285's own closing condition, unchanged, within 10 minutes of a release publishing and with nobody dispatching anything by hand:

```bash
curl -s "https://pdoom1.com/data/version.json?cb=$(date +%s)" | jq -r .latest_release.version
gh release view --repo PipFoweraker/pdoom1 --json tagName --jq .tagName
```

Until that passes unattended, treat a correct-looking pdoom1.com as evidence that somebody dispatched a workflow, not as evidence the pipeline works.


PipFoweraker:
Pushed `aee4978b` -- regenerated `/metabolism/` so the "Metabolism map is in step" check passes. The check failed correctly: the trigger block grew (cite `auto-update-data.yml` L9 -> L16) and a derived count moved.

**That derived count is the best evidence in this PR that the change does what it says**, because nothing asserted it:

```
- ...can RECEIVE repository_dispatch on 5 live workflow(s) (Bug Report Intake, Sync Design Tokens..., Sync Events..., Sync design notes..., Update Game Status)
+ ...can RECEIVE repository_dispatch on 6 live workflow(s) (Auto-Update Data, Bug Report Intake, ...)
```

Your own generator read the workflow files and counted a new receiver.

---

## One thing regenerating surfaced, which is yours and predates this PR

**The workflow table on `/metabolism/` carries a false claim, and check #5 on the same page contradicts it.**

Every scheduled repo-automation row renders the effect cell as:

> commits to the repo; the commit does not itself trigger a deploy

That string is applied uniformly, including to **Board liveness (score API)** -- whose completion is precisely what fires `auto-deploy-on-push` and rsyncs `public/`. Further down the same page, check #5 says the opposite, and says it correctly:

> **Automated commits reach production on their own -- YES.** ...auto-deploy-on-push.yml triggers on the 6-hourly board-liveness run, so public/ is rsynced ~4x a day whoever committed.

`generate-metabolism.py:1345` already documents the correction ("CORRECTED 2026-08-04... That was true when written and is now false"). The check was updated; **the table cell it shares a page with was not.**

So a reader is told both things. Narrowly the cell is defensible -- *that* workflow's own push does not trigger a deploy -- but on a page whose purpose is to tell someone what reaches production, "does not itself trigger a deploy" next to a workflow that is the deploy trigger reads as a safety guarantee that is not there.

**This bit me directly, which is why I am flagging it rather than filing it quietly.** I spent last night reasoning about whether committing `version.json` was safe unattended, and the estate's own documentation of that question says both yes and no in the same document.

I have not touched it -- it is a one-line change in your generator and I would be guessing at which reading you want. If you tell me the wording you prefer I will send it as a separate PR.

## Still deliberately untouched

Repo-root `data/current-game-version.json` -- zero consumers, outside the rsync source. Named in the PR body, not fixed. Yours.

## Correction to my PR body

The body says the manual two-step is what this PR deletes. Worth stating plainly since I got the surrounding facts wrong overnight and corrected them on `coordination#40`: **pdoom1.com was already serving v0.14.1 by ~02:26 AEST**, roughly 18 minutes after publish, because a seat ran Auto-Update Data and the deploy by hand. Nothing was broken for users. The argument for this PR is that a person did that work, not that the site was stale.

PipFoweraker:
## Unblocked: `main` merged in, and the concurrency question is answered

This PR was `CONFLICTING` / `DIRTY`. Two things were wrong, and **one of them the text merge would have resolved the dangerous way by itself.**

### 1. The concurrency group — resolved to `production-deploy`

This branch proposed the group name **`deploy-production`**. While it was open, **#299** ("one concurrency group across all FOUR production deploys", merged 2026-08-10) landed the same guard in `auto-deploy-on-push.yml` and its three siblings under the name **`production-deploy`** — the same two words, reversed — carrying this comment:

> `# ALL FOUR PRODUCTION DEPLOYS SHARE THIS GROUP. Do not rename it in one file.`

**The merge did not flag this.** The two blocks sit in different parts of the file — #299's above `on:`, this branch's below `permissions:` — so they conflict textually with nothing and git merged both. The result was a **duplicate `concurrency:` key**, with `deploy-production` last:

```yaml
concurrency:            # line 17, from #299
  group: production-deploy
...
concurrency:            # line 68, from this branch
  group: deploy-production
```

The last key wins. `auto-deploy-on-push` would have serialised against **nothing** — no longer with the other three, and no longer with itself under the name the other three use — while looking fully guarded in the file and in the Actions UI. That is #299's race fix silently defeated for one of the four deploys, and it is the failure mode that leaves no red square anywhere.

Resolved by **keeping `production-deploy`** and deleting the second block. This PR's body explicitly left the call to the reviewer:

> The concurrency group is deliberately incomplete … I did not make those edits because I do not know which of them you consider live.

#299 answers it: **all four are live, and they share one group.** The question is closed, not open.

Counts after the merge, across `.github/workflows/`:

| Spelling | Occurrences as a group value |
|---|---|
| `group: production-deploy` | **4** — `auto-deploy-on-push`, `deploy-dreamhost`, `version-aware-deploy`, `weekly-deployment` |
| `group: deploy-production` | **0** |

The only surviving `deploy-production` string in the repo is inside the replacement comment, which explains why it must not come back. Every workflow file parses with exactly one `concurrency:` key.

**Nothing else about the proposal changed.** `repository_dispatch: types: [game_version_sync]` on Auto-Update Data, the cron kept as the backstop, and `"Auto-Update Data"` added to the deploy `workflow_run` list are all exactly as proposed.

### 2. Stale `/metabolism/` — regenerated

The `check` job was failing `STALE: public/metabolism/index.html no longer matches what its sources produce`. Both diffs are the page correctly tracking this PR:

- The dispatch comment block sits above the cron, so `auto-update-data.yml`'s **cited line moved 9 → 16**. `/metabolism/` cites sources by `file:line`, so the fact is unchanged and only its coordinates moved.
- The new trigger makes this the **6th** workflow that can receive a `repository_dispatch` (was 5) — Auto-Update Data now appears in that list.

`python scripts/generate-metabolism.py --check` → `OK: /metabolism/ is in step with its sources.`

### Guard suite

All blocking Python checks from CLAUDE.md pass, including `check-stale-facts.py --min-severity HIGH`, `check-deploy-excludes.py`, `check-encoding-safety.py`, `validate_data.py` and `check-epoch-drift.py`.

Four checks fail — `check-stale-facts` (full report), `snapshot-copy --check`, `check-blessing-consistency`, `check-token-drift`. Those are **exactly** the four CLAUDE.md labels `[ADVISORY]`, and all four **fail identically on `main` at `cf30a8f3`**, so none is attributable to this branch. Node-based checks were not run locally (no node in that environment); CI covers them.

Merged `main` at `cf30a8f3` rather than rebasing, so review history is intact and no force-push was needed. **Not merged** — still yours to accept, amend or reject.

PR #291, opened 2026-08-08
fix(league): publish the L4 board, archive the closed L3 one

PipFoweraker:
pdoom1 **v0.14.0** forked the ladder **L3 → L4** on 2026-08-07. The site was still publishing `(weekly-2026-w31, L3)` — a closed epoch — as if it were this week's league.

Authority for the new key: the v0.14.0 git tag message, *"Ladder L4. Featured seed weekly-2026-w32. Board key (weekly-2026-w32, L4)."*, corroborated by `release_manifest.json` `"ladder_version": "4"`. Pip has blessed `(weekly-2026-w32, L4)`.

## What changed

**1. The outgoing board is archived first.** `public/leaderboard/data/preserved/2026-08-08-l3-epoch-close/` holds a read-only capture of `(weekly-2026-w31, L3)` and its 6 real entries, plus a README explaining that they are closed by an epoch cut, not lost. Without this, the instant the published key moves, `check-board-liveness.py` classifies that populated board as a **NEW ORPHAN** and goes red hourly about something everyone agreed on. Same mechanism and same reasoning as the `2026-08-01-gate4-proving-runs/` capture next door.

**2. `board-probe-targets.json`.** `current_ladder_epoch.value` → `L4`, sourced to the tag message + manifest + a positive observation; `declared_on` → 2026-08-08. `weekly-2026-w32` pinned under `extra_seeds` with a `source` and a `why_needed` in the same shape as the `weekly-2026-w31` entry — the probe derives seeds only from files that have already seen them, so a freshly drawn seed is invisible to it until a publish writes it down. That blind spot has now cost us twice, which is noted in `remove_when` as an argument for automating seed discovery rather than repeating this edit.

**3. A fourth `data_status`: `live-empty`.** The vocabulary was `live | pre-launch | legacy` and none of the three describes what a fork produces: the board key **is** current, the site **is** publishing the board players reach, and it holds no rows yet. `live` is the lie `validate_data.py` already fails on; `pre-launch` is a different lie ("the board is not on" when it is). `validate_data.py` accepts it **and fails it in the other direction** — `live-empty` with entries > 0 is also a FAIL, because a status that can never fire says nothing. The existing `live` + 0 entries FAIL is untouched. `ingest_scores.py` documents the shared vocabulary but never emits the new value: no file on disk can establish what the *current* key holds, so its empty case stays `pre-launch`.

**4. `publish-live-board.py` may publish an empty board, on a NEW key only.**
- same key + 0 rows → **refuse**. The key has not moved and the rows are gone; that is the API (or the network) losing data, and overwriting real scores with nothing is the damage this subsystem exists to prevent.
- new key + 0 rows → **publish**, as `live-empty`. Continuing to serve a closed epoch's standings as current is the worse lie.

"New" is measured against `leaderboard.json`'s own `meta.board_key` — what this site is serving right now — not against a second file's memory.

It still **never guesses a seed**. With zero rows the API is no evidence at all: every wrong key answers `ok:true` with an empty board. If several reachable boards on the epoch are empty, it refuses and names them. It narrows by *declaration* (a seed pinned with a `source`) and never widens.

`pick_live_board()` used to return `(None, None)` for both "no candidates" and "unreadable", disambiguated in `main()` by `3 if entries is None else 1`. It now returns an explicit status (`ok` / `unreadable` / `same-key` / `ambiguous` / `none`) and each refusal prints which one it is. **Test 2 — the transient-outage test — still asserts exit 3 and an untouched board.**

**5. Tests.** Old test 3 ("an empty board is never published") became **3a** same-key refuses, **3b** new-key publishes as `live-empty` with honest zero counts, and **3c** several empty boards → refuses and names them, then resolves once one is pinned with a source. The module docstring's safety-property list was rewritten to match, including a paragraph on *why* property 3 changed shape — the blanket rule was itself a way to lie.

**6. `public/leaderboard/index.html`.** `applyDataStatus` hides the search/sort/filter toolbar for `live-empty`, for the reason already written in the code for pre-launch: it "only offers a visitor ways to filter nothing". The banner says the board is open and waiting for its first finished run. **No new escaper**; the new string is a static template literal with no interpolation.

**7. `docs/LEAGUE_SEED_LEDGER.md`** gains a section recording the fork, explicitly marked **NOT a blessing** — the L3 row is still unfilled and only a human can fill it — and listing what a human still owes.

## What I verified

`(weekly-2026-w32, L4)` was checked **positively**, read-only, on 2026-08-08: it holds **9 entries** from clients stamped `v0.14.0` and `v0.14.1`. That matters because `ok:true` alone proves nothing — I confirmed `(weekly-2026-w32, L3)` and `(weekly-2026-w31, L4)` both also return `ok:true` with an empty board. Nothing was POSTed; GET only.

```
python scripts/validate_data.py                    0 fail, 3 warn, 32 ok
python scripts/test-publish-live-board.py          all PASS (9 groups)
python scripts/test_ingest_scores.py               PASS
node   scripts/test-board-honesty.js               PASS
node   scripts/test-board-escaping.js              OK
node   scripts/test-escaping.js                    OK
python scripts/check-stale-facts.py --min-severity HIGH   PASS (244 findings, 0 HIGH)
python scripts/check-encoding-safety.py            OK (60/63, 3 pre-existing waivers)
node   scripts/test-navigation.js                  14 passed, 0 failed
```

## What I did NOT do

- **`leaderboard.json` and `published-board.json` still name `(weekly-2026-w31, L3)`.** I was told not to run the publisher against the live API, so the actual switchover is left to the scheduled `board-liveness` / publish run — which is now unblocked by the declaration and the pin. Until it fires, the site keeps serving the closed board.
- The `validate_data.py` `board:key` check still reads the epoch from `board-liveness.json` (`L3`, written by the last bot run), so it currently reports "matches L3". It will follow once that workflow runs.
- No blessing was written. `ladder-epochs.json`, `seed_provenance.blessed`, and the ledger's L3/L4 rows are a human's to fill.

## Unsure about

- **The `live-empty` name.** It is my invention; `open-empty` or `awaiting-first-run` would do the same job. It is now in three files plus a data contract, so renaming later is cheap but not free.
- **The ambiguity rule is conservative and may bite.** In the real repo, `derive_targets` yields dozens of seeds, so if the L4 board had actually been empty, the publisher would have found many empty boards, narrowed to the two *pinned* ones (`w31`, `w32`) and **refused as ambiguous** rather than publishing. That is deliberate — with no rows there is nothing to distinguish them — but it means the empty-fork path only fires cleanly once the stale `weekly-2026-w31` pin is removed per its own `remove_when`. I left the stale pin in place rather than delete it, because its `remove_when` says to drop it after the next successful publish and that has not happened yet.
- **The premise "the fork board opens empty" was already false when I checked** — `(weekly-2026-w32, L4)` had 9 entries. So the populated path wins today and the new empty-publish code will not execute on the next real run. It is still the right fix (it was empty for several hours, and the next fork will do this again), but it ships mostly untested against production, with only the forced tests behind it.
- **`archived_keys()` matches on filename** via `^(?P<seed>.+)__(?P<version>[^_]+)\.json$`. I matched the existing convention exactly, so the archive entry should register — but I could not confirm it end-to-end without running the liveness probe, which writes `board-liveness.json`.


PipFoweraker:
**CI note — the one red check is pre-existing on `main`, not from this PR.**

`Metabolism map is in step / check` fails with `STALE: public/metabolism/index.html no longer matches what its sources produce`. Verified in a clean worktree of `origin/main`: it fails there too (exit 1). I then ran the generator on both `origin/main` and this branch and diffed the two outputs — **byte-identical**, so nothing here moved a cron, a park notice or a cited line.

Left unfixed on purpose: regenerating `/metabolism/` would put unrelated churn in a league-fork PR, and the fix (`python scripts/generate-metabolism.py`, committed) belongs on `main` where the staleness was introduced. Every other blocking check passes.

Last green run of that workflow on main was 2026-08-07 (`31158243629`), so the drift entered with one of the commits after it.

PR #295, opened 2026-08-09
feat(ci): an acknowledgement is a state with a clock, not a permanent exemption (class 5)

PipFoweraker:
Prototype answer to **"class 5, the knowing allowlist"** (`coordination#47`, workshop 2026-08-09):

> A check SEES the divergence, PRINTS it, and exits 0 **by design**. It is not disarmed, not mis-aimed, not stale. **The check is not fooled; the reader is, by its exit code.**

## The rule

**An acknowledgement is a state with a clock, not a permanent exemption. The thing that expires is the ACCEPTANCE, never the finding.**

- **Before `review_by`** — GREEN, and every acknowledged item is printed *and counted in the summary line*. Green carries a number, never silence.
- **After `review_by`** — RED, on *"this acceptance expired, re-accept or fix"*, **not** on the underlying finding. That red is always closeable by a human decision, which is what stops it rotting into the permanent red CLAUDE.md forbids. A red on the *finding* would not be — whoever hits it usually cannot fix it.
- **STALE** (key no longer fires) and **EXPIRING SOON** (inside the warning window) are both reported, neither blocking.

## What this found on day one

All three `KNOWN_UNFIXED` waivers in `check-encoding-safety.py` read *"held by the `<X>` branch (2026-07-29 sweep)"*. **Two of those three branches do not exist on the remote.** The justification had been false for days; the check went on printing `WAIVED` and exiting 0, because a reason without a clock cannot expire. That is the whole thesis, live in the repo.

I did **not** restate the dead reason. Each migrated entry says `LAPSED JUSTIFICATION` in its `why`, `accepted_by` says *"unattributed — migrated from the dict literal"* rather than forging Pip's name, and each gets a different short `review_by` (2026-08-23 / 08-30 / 09-06) so they do not all come due at once.

## Files

| file | what |
|---|---|
| `scripts/acknowledgements.py` | the module — no repo-specific logic |
| `data/acknowledgements.json` | the ledger; **repo root, not `public/`** (that is rsynced to prod) |
| `scripts/check-encoding-safety.py` | `KNOWN_UNFIXED` dict literal → the ledger |
| `scripts/test-acknowledgements.py` | **75 assertions, every state forced** |
| `.github/workflows/encoding-safety.yml` | test wired *before* the sweep, ubuntu + cp1252 Windows |
| `docs/decisions/ACKNOWLEDGEMENT_CLOCK.md` | portable design note for `pdoom1` / `pdoom-data` |

## Why `check-encoding-safety.py` as the reference

Small (3 entries), stable file-path keys, **already blocking-and-true in a real workflow** (so the clock lands on a gate that means something, not an advisory nobody reads), its allowlist was a **dict literal in a script** which CLAUDE.md's data-file rule forbids anyway, and its reasons had already lapsed so the mechanism had something true to say immediately.

Rejected, with reasons in the design note: `check-stale-facts.py` (its `SKIP_FILES` are structural facts about file *kinds*, not dated decisions); `snapshot-copy.py --check` (a review aid by design — clocking prose drift is a product call, not an engineering one); `test-header-consistency.js` (JS — porting means a second implementation).

## Refusal, not skipping

The loader raises and the check exits **2** on: missing/unparseable ledger, any missing or blank required field, non-ISO date, `review_by <= accepted_on`, an undeclared `check` name, a duplicate `(check, key)`, a `policy` value with no `source`.

A malformed entry is **not** treated as absent. "Absent" means "not acknowledged" means the check fails on the *finding* — which reads as a fresh bug and sends someone hunting something nobody introduced. Test 12 forces exactly this and asserts the run does not report the three real files as new findings.

## Guards run (verbatim tails)

```
$ python scripts/check-encoding-safety.py --self-test
self-test OK: 19 cases                                            exit=0

$ python scripts/check-encoding-safety.py --check
Acknowledgements for check-encoding-safety: 3 on file, 3 live and firing, 1 expiring within 14d
OK: 62 of 65 Python modules are encoding-safe (3 acknowledged, listed above)   exit=0

$ python scripts/check-encoding-safety.py --as-of 2027-01-01
FAIL: 3 acknowledgement(s) expired. Every encoding finding is either fixed or
acknowledged -- what is red is the ACCEPTANCE, listed above with what to do about it.
                                                                  exit=1

$ python scripts/test-acknowledgements.py
75 passed, 0 failed
All acknowledgement states forced and observed.                   exit=0

$ PYTHONIOENCODING=cp1252 python scripts/test-acknowledgements.py
75 passed, 0 failed                                               exit=0

$ python scripts/test-design-notes.py            26 passed, 0 failed          exit=0
$ python scripts/check-platform-claims.py        No unavailable platforms. OK. exit=0
$ python scripts/test-platform-claims.py         OK                            exit=0
$ python scripts/test-changelog-structure.py     All changelog tests passed!   exit=0
$ python scripts/check-stale-facts.py --min-severity HIGH
247 finding(s); 0 at or above the HIGH gate. PASS                  exit=0
$ python scripts/validate_data.py                0 fail, 3 warn, 32 ok         exit=0
$ node scripts/test-escaping.js                  OK                            exit=0
$ python scripts/check-deploy-excludes.py        PASS                          exit=0
```

`validate_data.py` rewrites `public/data/integration-health.json` as a side effect; reverted, not committed. Its 3 warns are pre-existing (seed version census).

---

## What I was unsure about, and what I think is wrong with the design

Stated plainly, as asked.

**1. Bulk re-accept is not prevented, only made visible.** Nothing stops a human bumping every `review_by` by ninety days without thinking. The design turns that into a dated, named, sourced diff in a reviewed file rather than making it impossible. I do not think a checker can make it impossible — it is a review norm. This is the design's weakest joint and I would not claim otherwise.

**2. The red lands on whoever pushes next, not on the accepter.** `EXPIRING SOON` and printing `accepted_by` make it routable, not solved. The proper fix is a scheduled job that opens an issue at `review_by - warn_within_days`; I did not build it because this repo has a documented history of workflows that open issues failing silently on missing `permissions: issues: write`, and it is a separable increment.

**3. I made STALE non-blocking, which is a judgement call you may want to overrule.** Your brief said a vanished finding "should also be surfaced" without saying whether it blocks. I chose loud-but-green: deleting a dead exemption is housekeeping, not an honesty risk *today*, and its own `review_by` forces the question on a date anyway. Making it blocking would create a permanent red for tidiness — the exact failure mode. But a stale entry **does** still block once past `review_by` (test 4b), so it cannot silt up forever.

**4. Key stability is an unguarded assumption.** If a check's key format changes — a renamed file, a reworded finding — every acknowledgement silently becomes `STALE`. Visible, but as the *wrong diagnosis*: it reads as "the finding went away" when the truth is "the check stopped speaking the ledger's language". The `checks` map catches a renamed **check**; nothing catches a renamed **key**. I do not have a clean answer.

**5. `warn_within_days: 14` is invented.** It carries a `source` note saying it is a judgement call and not derived from measurement, which is the honest form, but it is still a number I made up.

**6. The migrated entries record an absence of authority.** `accepted_by` says "unattributed". Every field validates, so the ledger accepts an entry whose accepter is a paragraph explaining there is no accepter. That is deliberate — forging your name would be worse — but it does mean the required-field check can be satisfied with an honest non-answer. Field presence is enforceable; field *meaningfulness* is not.

**7. The real fix here is three lines of preamble, and I did not apply it.** Adding the preamble to `ingest_scores.py`, `sync/sync-events.py` and `weekly-league-manager.py` would empty the ledger and delete the problem. I left it because it touches three scripts with their own tests and would have left the reference implementation with no data to demonstrate against. **Consequence you should agree to before merging: this check goes red on 2026-08-23 unless someone acts.** That is the mechanism working, not a defect, but it is a commitment being made on someone's behalf and you should say yes to it explicitly.

**8. A small CLAUDE.md conflict I resolved by judgement.** CLAUDE.md argues against a shared helper import for the cp1252 preamble ("an import that raises defeats the exact purpose"). `check-encoding-safety.py` now imports `acknowledgements` via a `sys.path.insert`. I judged this different in kind — it is not the preamble mechanism, `acknowledgements.py` carries its own preamble, and it imports only stdlib — but it is a new way for that check to fail to start, and you may disagree.

**9. One thing I think is right about your design and want to name**, since the rest of this section is objections: making the red be about the *acceptance* rather than the *finding* is the move that makes this work at all. I tried to find a version where the red is about the finding and it always collapses back into red-forever, because the person who hits it has no way to close it. The distinction is not cosmetic.


PipFoweraker:
## Pip signed off verbally, 2026-08-12 ~10:15

Capture `2026-08-12_101456__94ddce73`, transcribed on-device. Verbatim:

> *"E1, which is asking me to merge PR295, the acknowledgement clock... merging it commits me to a red on check unless something acts by then. Or, clear the three waivers instead... **No, I am committing myself to a red check on this by the 23rd of August unless someone acts by then. That's the mechanism acting as designed, but it's a commitment made on my behalf. That's fine, signed verbally as of now.**"*

He was offered both options the pack put to him — accept the 2026-08-23 commitment, or clear the three waivers with a preamble and empty the ledger — and **chose the commitment, knowingly.** He named the consequence in his own words before accepting it.

`PACK_2026-08-11_review-284-and-295.md` held this out of the safe-to-merge pile precisely because it has a consequence. That condition is now discharged by the person it binds.

**Merging.** The 2026-08-23 review date is live from today. Three entries come due on 08-23, 08-30 and 09-06 respectively, deliberately staggered so they do not all fire at once.

Recorded by the `AlsoBort` seat acting as coordination.

PR #306, opened 2026-08-11
fix(parsers): harden the two real findings from the delimiter sweep

PipFoweraker:
Follow-through on the 2026-08-11 sweep, which attacked four parsers with adversarial input. **Two were genuinely broken. One was fine and I was wrong about it.**

## 1. The seed comes from its own cell — real, mine, shipped in #301

It took *"the first backticked value in the row that is not a filename"*. Reads sensibly; wrong:

```
| L9 |  | L9 |  | - | - | see `docs/THING.md` for the seed |
       ->  seed extracted: 'docs/THING.md'
```

**A documentation path presented as a blessed league seed.** Harmless today only because the live unfilled row is flagged separately and nulls the seed first — an accident, not a guard. That is the defect class that once stranded 23 submissions.

Two regression tests, and **the second one found a further gap on its own**: a ledger row that claims a blessing while naming **no seed** used to pass silently. It now reports — a blessing is a sign-off on a specific string, and without one there is nothing for the other artefacts to agree with.

## 2. A `>` inside a quoted attribute does not close a tag

`sync-keybinds` used `[^>]*` for the attribute run, which stops at the first `>` regardless of quoting:

```
<kbd data-keybind="move>right">X</kbd>
   attrs -> ' data-keybind="move'    inner -> 'right">X'
```

It would look up a truncated action, fail to substitute, and leave the placeholder standing — **sending a player to a key that does nothing, which is the exact outcome this mirror exists to prevent.** Verified against both quote styles and normal input; the real check still passes.

## 3. Metabolism now says *which* class of stale

The page cites sources by `file:line`, so inserting anything above a citation moves the anchor **without any fact changing**. That printed the same words as a genuine cadence change, and it cost me twenty minutes twice in four days.

- **STALE (COORDINATES ONLY)** — a cited line moved, nothing changed, regenerate
- **STALE (CONTENT)** — a cadence genuinely changed, **read the diff first**

## Probe 4 is WITHDRAWN and deliberately not actioned

I claimed the `<script>` stripper ended early at a `</script>` inside a JS string and silently lost coverage. **Checked against a real HTML parser: it does not.** Per spec that sequence genuinely terminates the element — which is why you write `<\/script>` in JS. The regex matches browser behaviour and my "false negative" reasoning was backwards; ending early scans *more*.

Three probes were attacks. The fourth was an argument — and it was the one I called *"the most interesting"*. Correction is in `coordination`'s pack plus a printed addendum.

## Verified

`test-blessing-consistency` (11 states) · `generate-metabolism --check` · `sync-keybinds --ci` · `check-encoding-safety` · `test-epoch-drift` — all green.

### DIFF
--- a/scripts/check-blessing-consistency.py
+++ b/scripts/check-blessing-consistency.py
@@ -112,13 +112,29 @@ def parse_ledger(path):
         epoch, rest = m.group(1), m.group(2)
         cells = [c.strip() for c in rest.split("|")]
         unfilled = bool(UNBLESSED_RE.search(rest))
-        ticked = BACKTICKED.findall(rest)
-        seed = None
-        for t in ticked:
-            # the seed cell is the first backticked value that is not a filename
-            if not t.endswith(".json"):
-                seed = t
-                break
+
+        # THE SEED COMES FROM ITS OWN CELL, and from nowhere else.
+        #
+        # This used to take "the first backticked value in the row that is not a
+        # filename", which reads sensibly and is wrong: a row whose seed cell is
+        # EMPTY adopts the first backtick anywhere in the row, including a
+        # documentation path in the notes column --
+        #
+        #   | L9 |  | L9 |  | - | - | see `docs/THING.md` for the seed |
+        #        -> seed == 'docs/THING.md'
+        #
+        # which would present a doc path as a blessed league seed. Found by
+        # attacking the parser (the 2026-08-11 sweep), not by reading it. It was
+        # harmless only because the live unfilled row is flagged separately and
+        # nulls the seed before this runs -- an accident, not a guard.
+        #
+        # Cell 0 is the seed column: ROW_RE has already consumed the epoch cell,
+        # so `rest` begins at the seed. Refuse rather than guess when it is empty.
+        seed_cell = cells[0] if cells else ""
+        ticked = BACKTICKED.findall(seed_cell)
+        seed = ticked[0] if ticked else (seed_cell or None)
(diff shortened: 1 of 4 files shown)


PipFoweraker:
## §5g RED run, observed and recorded

`RULED_2026-08-10` R1 lets a seat merge the guard class alone but does **not** exempt it from failing once first. This PR described its RED in prose — *"a ledger row that claims a blessing while naming no seed used to pass silently"* — without recording a run. Produced here instead, locally and reproducibly, so the guard counts as installed rather than claimed.

**Method:** take this PR's new `scripts/test-blessing-consistency.py` and run it against **main's un-hardened parsers**.

**RED** — new test, old parser:
```
10. THE SWEEP BUG: an empty seed cell must not adopt a backtick from the notes
  FAIL  a documentation path from the NOTES column is never read as a seed
1 FAILURE(S)
exit code: 1
```

**GREEN** — same test, this branch:
```
exit code: 0
```

Reproducible from a clean clone:
```
git checkout main && git checkout pr306 -- scripts/test-blessing-consistency.py
python3 scripts/test-blessing-consistency.py   # exit 1
git checkout pr306 && python3 scripts/test-blessing-consistency.py   # exit 0
```

No CI run id because the RED state never existed on a branch CI saw — the test and the fix arrived in the same commit. The reproduction above is the stronger artefact anyway: it can be re-run by anyone, on any machine, at any time.

**Self-merged by the `AlsoBort` seat under R1, guard class**, recorded here per R1 condition 2.

PR #284, opened 2026-08-07
content: v0.14.0 announcement, its campaign draft, and an art drip that posts verdicts

PipFoweraker:
Three drafts. **Merging this publishes the blog post** — the other two files sit outside `public/` and reach no reader.

> ## CORRECTED 2026-08-13 — read this before merging
>
> Rebased onto `main` (was 143 behind; clean, no conflicts). Three corrections to
> the body below, found by checking it against the diff:
>
> **1. The campaign file is `content/campaigns/2026-08-07-ladder-fork-l4.json`,
> not `2026-08-07-v0-14-0-release.json`.** Section 2 named the pre-rename path.
> Commit `4202839a` ("key the copy to the fork, not to a version number") renamed
> it, for the reason the file's own `_why_the_slug_has_no_version_in_it` records:
> a `utm_campaign` value is permanent once posted.
>
> **2. FIXED — the "step behind" section was present tense about a solved
> problem.** Section 1 below sells that paragraph as a virtue ("better said by us
> than discovered by a player"). That was accurate when drafted on 08-07 and is
> not now: `check-epoch-drift.py` reports `"verdict": "in-step"` — site declares
> `L4`, pins `weekly-2026-w32`, published epoch `L4`, and the shipped client
> (manifest `v0.14.1`) posts to `(weekly-2026-w32, L4)`. Publishing it unedited
> would have told a visitor the league page may be showing a superseded board
> when it is not — a fresh false claim, made by the one section whose whole
> purpose is candour.
>
> Pip reviewed this copy on 2026-08-12 and approved it conditionally on exactly
> this change; `SHEET_2026-08-12_pr-decisions.md` records the same instruction.
> Commit `618beff5` makes it, **tense only** — the window still happened, it is
> still disclosed, and the reason is unchanged. The heading moved too, because
> "The site is a step behind" asserts the condition on its own. The front-matter
> `summary` carried the phrase as well and is the field that propagates, so
> `index.json`, `feed.xml` and `atom.xml` were regenerated rather than hand-edited.
>
> **3. The licence correction landed in the blog post but not in the campaign
> copy.** Commit `5bb623f8` changed the post's footer to "free, **source-available**
> and non-commercial", because `/press/` had that claim removed in #310 on the
> grounds that the interim licence grants no redistribution. The campaign JSON on
> this same branch still says "Free, **open source**, non-commercial" (LinkedIn),
> "Free, **open source**, still an alpha" (Instagram), and carries
> "The game is free, **open source** and non-commercial" as a
[...]

PR #307, opened 2026-08-11
fix(analytics): Linux download event no longer names an unpublished asset

PipFoweraker:
## What this fixes, precisely

Related to PipFoweraker/pdoom1#1068 ("The site's Linux download button 404s (PDoom.x86_64 is not a release asset)").

`public/index.html`'s `setupDownloadTracking()` had a `TARGETS` array used to build the Plausible `Download` event's `file` prop:

```js
{ id: 'download-linux', platform: 'Linux', file: 'PDoom.x86_64' }
```

**Verified against the live v0.14.1 release assets** (`gh release view v0.14.1 --repo PipFoweraker/pdoom1 --json assets`): `PDoom-Windows.zip`, `PDoom.app.zip`, and `PDoom-Linux.zip` all exist as real, downloadable assets. `PDoom.x86_64` does not — it's the executable packed *inside* the Linux zip, not a published release asset, and `curl`-ing its would-be direct URL confirms a 404.

## Important scope correction from my investigation

Before editing, I traced how `TARGETS` is actually used, since #1068's cited line/behavior didn't match what's currently in `public/index.html`. Two things worth being explicit about:

1. **The Linux button's actual click-through `href` is not sourced from `TARGETS` at all.** It starts as the release-page baseline (`.../releases/latest`) and is upgraded at page load by `resolveDownloads()`, which fetches the live release's real asset list from the GitHub API and regex-matches each platform's actual filename. That mechanism (added in commit `493fac55`, well before #1068 was filed) already handles arbitrary/versioned asset names correctly — for v0.14.1 it resolves the Linux button to the real `PDoom-Linux-v0.14.1.zip` asset, a working link.
2. **`TARGETS` is consumed only by the Plausible analytics `file` prop** (see `setupDownloadTracking()`), fired on click. So the practical effect of the stale `PDoom.x86_64` value was silently wrong analytics data — every Linux download to date was logged under a filename that was never published — not a dead link for visitors clicking the button today.

I'm not certain whether #1068 was observing an older deploy or reasoning from a static grep of the file without accounting for `resolveDownloads()`'s dynamic override. Either way, this PR fixes the one real, currently-live defect I could confirm: the Linux entry now matches the stable unversioned asset name (`PDoom-Linux.zip`), the same pattern Windows and macOS already use, so the analytics `file` prop reflects a real, resolvable asset.

**Windows and macOS entries were already correct and are untouched by this change.**

## Diff

One line changed in `public/index.html`:
```diff
- { id: 'download-linux',   platform: 'Linux',   file: 'PDoom.x86_64' }
+ { id: 'download-linux',   platform: 'Linux',   file: 'PDoom-Linux.zip' }
```

## Test plan

- [x] `gh release view v0.14.1 --repo PipFoweraker/pdoom1 --json assets` confirms `PDoom-Windows.zip`, `PDoom.app.zip`, `PDoom-Linux.zip` all exist as real assets.
- [x] Confirmed `scripts/test-download-resolution.js` (the CI check covering download-button behavior, wired into `content-honesty.yml`) only exercises `resolveDownloads()`, not the `TARGETS` array touched here — this change is outside that test's surface, and I could not run it locally (no `node`/`npm` in this environment) but read it in full to confirm non-interference.
- [x] Grepped `scripts/check-stale-facts.py` and `scripts/check-platform-claims.py` for any reference to `TARGETS`/`x86_64`/`file:` — none found, so no other guard is affected.
- [ ] Pip to verify in the Netlify deploy preview and confirm intended scope (analytics-accuracy fix, not a click-through fix, since the click-through was already working).

**Do not merge — flagging for Pip's review per standing instruction on anything touching the public site.**

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1760,7 +1760,7 @@ t	.hero {
 			const TARGETS = [
 				{ id: 'download-windows', platform: 'Windows', file: 'PDoom-Windows.zip' },
 				{ id: 'download-macos',   platform: 'macOS',   file: 'PDoom.app.zip' },
-				{ id: 'download-linux',   platform: 'Linux',   file: 'PDoom.x86_64' }
+				{ id: 'download-linux',   platform: 'Linux',   file: 'PDoom-Linux.zip' }
 			];
 
 			TARGETS.forEach(function(target) {


PipFoweraker:
## Merge-order hazard: this branch reintroduces a licence claim that #310 removes

`public/blog/2026-08-07-v0-14-0-the-ladder-forks.md:94` reads:

> *"p(Doom)1 is free, **open source** and non-commercial."*

That contradicts `public/index.html:1226`, which states deliberately and emphatically: *"Open source: **not yet, and it would be wrong to say otherwise**. …the current interim licence does not grant redistribution."*

**#310** corrects three instances of this in `public/press/index.html` — but **this blog post is not on `main` and not on #310's branch**, so #310 cannot fix it. Measured:

| branch | has the post | "open source" hits |
|---|---|---|
| `main` | no | – |
| `content/v0-14-0-announcement` (#284) | yes | 1 |
| `fix/linux-download-asset-name` (#307) | yes | 1 |
| #310 | no | – |

**So whichever of #284 or #307 merges first publishes the false claim onto `main`, after the fix has landed.** One word: `open source` → `source-available`.

PR #314, opened 2026-08-11
ci(guard): make class:guard and class:docs actually check something

PipFoweraker:
## What this is

Ruling R1 (`RULED_2026-08-10_pr-self-merge-and-four-more.md`) lets a seat merge
the **guard** and **docs** classes without Pip, on green. The labels `class:guard`
and `class:docs` were created in three repos on 2026-08-10 and, right now, they
**promise eligibility and nothing checks anything** -- the same defect that
retired `ship:hotpatch-48h` the same day. This PR is the mechanism, ported from
`pdoom1#1203`.

New workflow `.github/workflows/self-merge-eligibility.yml` (pull_request:
opened, labeled, unlabeled, synchronize, edited), backed by
`scripts/check-self-merge-eligibility.py` and 47 hermetic assertions in
`scripts/test-self-merge-eligibility.py`. Stdlib only, no pip install.

| situation | verdict |
|---|---|
| no class label | PASS, neutral -- never blocks a normal PR |
| `class:docs` | PASS only if EVERY changed path is internal documentation |
| `class:guard` | PASS only if the PR body records the RED run (Section 5g) |
| `needs:pip` + a class label | FAIL -- a label cannot lift a hold |
| both class labels | FAIL -- a PR is one class or neither |

**What "documentation" excludes here, and why.** Everything under `public/` is
served to visitors, so editing it is changing a **public claim** -- which R1
keeps with Pip and which the prime directive makes the highest-stakes edit in
this repo. A published blog post is a `.md` file and still fails the docs class.
Also excluded: `docs/copy-baseline/` (editing the frozen snapshot is how a
copy-drift check is made to agree with drifted copy), `deploy-excludes.txt`,
`requirements.txt`, `runtime.txt`.

## Section 5g: this PR is currently the deliberate RED

This PR was opened carrying `class:guard` with **no declaration in the body**, on
purpose, so the new gate would fail on this very PR for the exact rule it exists
to enforce.

RED-RUN: https://github.com/PipFoweraker/pdoom1-website/actions/runs/31491900189 -- opened with class:guard and no declaration in the body; the gate failed itself before it was believed

That run is red at `Check this PR against its class label`. The two steps above
it were green in the same run: the 14-case rule table and the 47 hermetic
assertions ran before the gate judged this PR. This line is what turns it green
-- so the check's own green is downstream of its own red.

~~DO NOT MERGE -- reporting back first.~~

**HOLD CLEARED by Pip, 2026-08-13 ~08:45: "merge 314".**

The line above was written while this PR was deliberately RED and was never retracted once it went green. Recording the retraction here rather than deleting the line, because the body of a merged PR is a durable record and this estate has already been bitten this week by one that said something untrue.

Verified before merge: `MERGEABLE`/`CLEAN`; the guard was forced to fail on this PR's own git-derived diff (RED-RUN line stripped -> exit 1; `class:docs` on a non-docs diff -> exit 1 naming all four paths; `needs:pip` + `class:guard` -> exit 1; no label -> neutral pass), so its green is load-bearing rather than vacuous.

Generated with [Claude Code](https://claude.com/claude-code)


PipFoweraker:
## Independent verification, 2026-08-13 — the guard is not vacuous

Rebased onto `main` (was 34 behind; clean). All checks green, including
`Self-merge class eligibility`.

**The §5g RED run is real, not just cited.** Run `31491900189` concluded
`failure`, and the failing step is `Check this PR against its class label` — the
exact rule this PR installs. The two steps above it (`Prove the gate can still
fail (self-test)` and `The rules hold one at a time, not just as a table`) were
`success` in that same run, so the gate's own green is downstream of its own red,
as the body claims.

**A green here could still have meant "the check never fires", so I forced it to
fail.** Not the hermetic `--self-test`, which proves only that the rule table
agrees with itself — the *real* entrypoint, with real git-derived paths from this
PR's own diff:

| what I ran against this PR's actual diff | exit | verdict |
|---|---|---|
| `class:guard` + the real body | 0 | PASS — declaration found |
| `class:guard` + body with the `RED-RUN:` line stripped | **1** | **FAIL** — the rule it exists to enforce |
| `class:docs` on this (non-docs) diff | **1** | **FAIL** — names all four offending paths |
| `class:guard` + `needs:pip`, real body | **1** | **FAIL** — a label cannot lift a hold |
| no class label | 0 | NEUTRAL — never blocks a normal PR |

So the passing state is load-bearing: remove the one line the guard is about and
it goes red on real inputs, not on a fixture.

Body checked against the diff and found accurate throughout — the `14`-case rule
table and the `47` hermetic assertions both reproduce locally (`14 case(s) ran, 0
mismatch(es)`, `47 passed, 0 failed`), stdlib-only holds, and the docs-class
exclusions match `NOT_DOCUMENTATION_PREFIXES` / `NOT_DOCUMENTATION` exactly. No
corrections needed.

Also re-run locally on the rebased tree: `generate-metabolism.py --check`,
`check-encoding-safety.py --self-test` and `--check`, `test-acknowledgements.py`,
`check-platform-claims.py`, `test-platform-claims.py`,
`check-stale-facts.py --min-severity HIGH`, `check-deploy-excludes.py`,
`check-published-emails.py`, `build-blog-index.py --check`,
`generate-feeds.py --check` — all pass.

**One transient to note, so nobody re-diagnoses it:** the Netlify deploy preview
failed once after the rebase. Rebuilt from a byte-identical tree (same tree hash,
new commit SHA) and it passed. Not caused by anything in this PR.

**Not merged.** Every R1 condition is met — guard class, green, §5g satisfied and
recorded — but this PR's body still ends `DO NOT MERGE -- reporting back first`,
written when it was deliberately RED. That line has not been retracted, and
clearing an author's own written hold is not a seat's call to make on inference.
This comment is the report-back it asks for; merging needs one word from Pip.

PR #315, opened 2026-08-12
fix(encoding): clear all three waivers by fixing the findings, and stop the tests pinning them

PipFoweraker:
Closes the three `check-encoding-safety` acknowledgements from #295 by **fixing the findings**, not by re-accepting them. The ledger now holds zero entries, and the sweep is green on all 73 modules **with nothing acknowledged**.

The soonest expiry (`sync-events.py`, `review_by` 2026-08-23) is cleared, so nothing goes red on Sunday.

## Waiver inventory

All three waivers live in `data/acknowledgements.json`, read by `scripts/acknowledgements.py`, consumed by `scripts/check-encoding-safety.py`, gated by `.github/workflows/encoding-safety.yml`. There were exactly three, matching the brief.

| Waived file | Cited justification | Branch exists? | Finding real? | Disposition |
|---|---|---|---|---|
| `scripts/ingest_scores.py` | "held by the **leaderboard** branch (2026-07-29 sweep)" | **No** — verified `git ls-remote --heads origin` | **Partly.** Wrapped `stdout` only; `stderr` stayed cp1252, so a traceback still died | Fixed, waiver deleted |
| `scripts/sync/sync-events.py` | "held by the **events-sync** branch (2026-07-29 sweep)" | **No** | **No.** Wrapped **both** streams under `win32` — fully protected on the platform the finding is about | Fixed (normalised), waiver deleted |
| `scripts/weekly-league-manager.py` | "held by the **league** branch (2026-07-29 sweep)" | Only `origin/league/epoch-boundary`, a *similar name*, never confirmed to touch the file | **Yes.** No guard at all, 58 `print()` calls | Fixed, waiver deleted |

The dead-branch half was already found and honestly recorded in #295 — each entry says `LAPSED JUSTIFICATION` in its own `why`. What #295 did **not** find is the second half below.

## Root cause: three separate mechanisms, and only one was known

**1. A reason with no clock.** The original `KNOWN_UNFIXED` dict literal carried a reason and no expiry, so "held by branch X" went on being printed for eleven days after branch X stopped existing. Already fixed by #295's acknowledgement clock. Not re-litigated here.

**2. The check tests one spelling, not the property.** `PREAMBLE_RE` matches `reconfigure(encoding="utf-8"` and nothing else. Two of the three files solved the Windows crash with `io.TextIOWrapper(sys.stdout.buffer, ...)`, which works, and the sweep reported them as unprotected anyway. That is how a waiver came to be written for `sync-events.py` asserting it was *"the one most likely to hit the crash on a Windows run"* — it was the one module of the three that could not, because it was the only one that had wrapped **both** streams. **A waiver inherits the check's blind spot and then adds prose that sounds like evidence.**

**3. The tests made the waivers load-bearing — this is the answer to "does anything ever remove a waiver".** `scripts/test-acknowledgements.py` asserted `len(led.entries) == 3`, named all three waived paths in case 13, and drove the expiry wiring off the live ledger. So the commit that *fixes* all three findings turns the test suite **red**. The suite punished precisely the outcome the ledger exists to produce. Nothing removes a waiver, because removing one breaks CI. An allowlist whose size is asserted is an allowlist nobody can empty.

## What changed

**The three findings, fixed** — canonical preamble verbatim, not platform-gated, in `ingest_scores.py`, `sync/sync-events.py`, `weekly-league-manager.py`. `data/acknowledgements.json` `acknowledgements` is now `[]`; the machinery, policy and `checks` map are untouched.

**The tests, so a waiver can be deleted without breaking CI** (cases 9–13):
- case 9 asserts the invariants for *whatever* is on file — unexpired, distinct `review_by`, cites a source — and no longer pins a count. It prints the count instead, noting zero is the goal state, not a broken fixture.
- cases 10 and 11 force expiry against a **synthetic** ledger. Driving them off the real one only worked while the real one happened to hold entries; once it was empty, `--as-of 2027` would have had nothing to expire and both would have gone **green for the wrong reason** — the worst outcome available to a guard.
- case 13 is inverted. It used to prove suppression was doing real work by emptying the ledger and watching the three findings go red. It now asserts that with **nothing** acknowledged the check is **still green** — which separates "no findings" from "findings, quietly forgiven", and goes red the day someone papers over a regression with a new waiver instead of a fix. That the scanner can still *fail* no longer depends on the repo being dirty; `--self-test` proves it (19 cases, its own CI step).

**`CR CR LF`, the other half of the brief.** `docs/archive/security-audit-report.json` was the last file in the estate with doubled carriage returns — 44 × `\r\r\n`. Reproduced the trap before fixing it: one `CRLF -> LF` pass, which is exactly what `git add --renormalize` does, leaves **44 lone CRs** behind and looks fixed. Removed every CR byte instead (90 of them). Verified against `HEAD` that the result is byte-identical once CRs are stripped from both sides — line endings only, zero content change.

`docs/SURFACE_REGISTER.md` carried a raw CR and a raw LF inside two markdown code spans where `\r` and `\n` were meant; the spans rendered as empty backticks. Same mangled-escape class `check-control-characters.py` was written for — and that checker allows CR by design, so nothing would ever have caught it.

**Docs.** CLAUDE.md's "Environment / tooling" still pointed at `KNOWN_UNFIXED` and claimed three files were held by concurrent branches. Corrected in place per this repo's own rule that a fix is not done until the paragraph moves. The class-5 narrative at CLAUDE.md:488 and `ACKNOWLEDGEMENT_CLOCK.md` are left alone — they are past-tense history and still true.

## How this was proved

Exit codes were not treated as evidence of anything.

**Line endings — bytes re-read from disk after every edit:**

```
BEFORE: bytes=2306  CRCRLF=44  CRLF=2  loneCR=0
after ONE CRLF->LF pass: loneCR left = 44   <-- the trap, reproduced live
AFTER : bytes=2216  CR(any)=0  CRCRLF=0  LF=46
CR-stripped identical to HEAD: True   (old CRs 90, new CRs 0)
```

A repo-wide rescan of every tracked text file across `pdoom1`, `pdoom1-website`, `pdoom-data` and `coordination` now finds **zero** `\r\r\n` anywhere.

**The preamble — forced failure, then the same forcing against the fix.** Under `PYTHONIOENCODING=cp1252:strict`, printing U+2713 raises `UnicodeEncodeError`. Loading each module and then printing it:

```
against main:        0/3 modules survived
against this branch: 3/3 modules survived  (stdout AND stderr now utf-8)
```

Re-running with `sys.platform` spoofed to `win32` is what separated the three cases in the table — it is how `ingest_scores.py` was caught protecting `stdout` and leaving `stderr` on cp1252. (`sync-events.py` cannot be executed that way on Linux: spoofing `sys.platform` breaks the stdlib `shutil` import with `No module named '_winapi'`. That crash is an artifact of the spoof, not a finding; its protection was read from source, where it wraps both streams unambiguously.)

**Suites run green:** `check-encoding-safety` (73/73, nothing acknowledged), `--self-test` (19), `test-acknowledgements` (75), `test_ingest_scores`, `test-weekly-league-boundary` (99), `test-sync-events`, `test-sync-events-pii` (31), `check-control-characters` (unchanged: warns only on generated event pages).

## Proposed, deliberately NOT implemented today

Code freeze, release tomorrow. Each of these is a real structural gap and none is small enough to land unreviewed:

1. **Make W1 test the property, not the spelling.** The honest fix is an AST check that `sys.stdout` and `sys.stderr` are both reconfigured or replaced on every path, rather than a regex for one snippet. Widening `PREAMBLE_RE` is the tempting cheap version and is worse — it would bless divergent idioms that CLAUDE.md deliberately forbids, and the whole point of prescribing a literal snippet is that it is greppable. Until then the checker now says so in a comment at the ledger site.
2. **Require a waiver's cited branch to be live.** Every one of these three named a branch, and two had evaporated. `source` is free text, so nothing can validate it. A `blocked_by_branch` field the loader could check with `git ls-remote` would have caught this on 2026-07-30 instead of 2026-08-09.
3. **Nothing checks that a test does not pin the allowlist.** Root cause 3 is the interesting one and it has no guard. Worth a rule in `ACKNOWLEDGEMENT_CLOCK.md`'s porting checklist before the clock is ported to `pdoom1` and `pdoom-data`, so the same trap is not rebuilt twice more.

## Deliberately left alone

- **`public/events/arxiv_407a89d50dd6ff35.html`** — 2 lone CRs, and 28 other control chars across 8 generated event pages. Generated from upstream PDF text; `check-control-characters.py`'s own severity model says the fix belongs in the generator. Tracked at pdoom-data#45. Untouched.
- **`coordination/knowledge/MAIN-full.md`** — 281 CRLF, consistent, **no** `\r\r\n`. `coordination` has no `.gitattributes` and declares no LF convention, so there is nothing here to be inconsistent with. Not this PR's repo either.
- **`pdoom1`** — clean. `394753f0` ("one renormalize pass was not enough") did finish the job; zero CRs in tracked text files, verified. No waiver mechanism exists there. Nothing touched, and nothing in `art_generated/`, `art_prompts/` or `tools/art_review/` was read or modified — the art review is live in that checkout, and all work here was done in a separate worktree.
- **`pdoom-data`** — `check_invariants.py`'s `KNOWN_COUNT_DIVERGENCES` is already `{}`. No encoding waivers. Clean.
- **The other three in-script allowlists** (`check-platform-claims.py`'s `ALLOWLIST`, `check-stale-facts.py`'s `SKIP_FILES` and `LINE_ALLOWLIST`) — the same class-5 shape, none of them encoding-related, and porting them to the clock is the migration `ACKNOWLEDGEMENT_CLOCK.md` already plans. Not a freeze-day change.

## A note on the earlier decision

The remediation option chosen for #295 was picked before the dead-branch problem was known, from a printout that was four hours stale. That choice — put a clock on the three waivers and let them run to `review_by` — is superseded by what is in the table above, not contradicted: two of the three files did not need a waiver at all, and the third needed five lines. Carrying a clock on findings that can be closed in one commit would have spent three review deadlines to arrive here anyway.


PipFoweraker:
CI green, including **Real scripts under a cp1252 stdout** (`windows-latest`, `PYTHONIOENCODING=cp1252`) — the only job in this repo that can actually observe the crash this PR is about. The Linux jobs cannot, by construction.

One follow-up commit, `7a9a3e16`. `/metabolism/` cites its sources as clickable `file:line`, and the four preamble lines shifted `get_current_week_info()` in `weekly-league-manager.py` from L308 to L312, so `generate-metabolism.py --check` went red. Regenerated: **one line** of the page changes, the citation link only, with no cadence, cron or park notice moved.

Worth noting as a small point in this PR's favour — the metabolism guard caught a citation going stale on the same day the line moved, which is the failure mode the rest of this PR is about, working correctly for once.

PR #327, opened 2026-08-17
DRAFT: feedback intake - open to make CI run the destructive suite against ingest.php

PipFoweraker:
**Draft. Not for merge.** Opened to get signal, not approval.

## Why this exists

`public/ingest.php` is the endpoint at the centre of `docs/decisions/FEEDBACK_INTAKE_CONTRACT.md` and **it has never executed anywhere**. There is no `php` binary on the dev box, there was no PR, and `feedback-intake.yml` triggers on `push` to `main` only. So `scripts/test-ingest-destructive.py` — thirteen well-written forced-failure tests covering F1-F6 and F9-F15 — has never once touched the thing it tests.

Every safety claim in that file's 92-line header is currently a docstring rather than evidence. This repo's own rule: *a guard seen only in its passing state has not been shown to work*, and this one has not been seen at all.

Opening the PR runs `feedback-intake.yml`, which installs PHP 8.2 via `shivammathur/setup-php@v2` and executes the suite. **The result is the point, red or green.**

## Expected to be RED, for known reasons

- **Gate 2 is deliberately red pre-implementation** — that is the contract's design.
- `check-prose-mechanism-coupling.py` exits 1: `public/privacy/index.html` promises a public tally with k=5 suppression and a deletion clock, and nothing generates or schedules either. Being fixed in a separate lane.
- Known defects under repair, found by review on 2026-08-17: `purge-feedback.py` takes no lock while `ingest.php` locks an inode that `os.replace()` swaps out (a submission POSTed during a purge can be acknowledged and then vanish); `feedback-maintenance.yml` treats a mistyped store path as green; `purge-feedback.py`'s docroot guard uses a name heuristic that misses DreamHost's real layout.

## What NOT to do with this PR

Do not merge it to make it green. Do not merge it at all yet. Read the `intake-tests` job output, and tell us which of F1-F15 actually hold.


PipFoweraker:
## Answering the question this PR was opened to ask

Run [32001215197](https://github.com/PipFoweraker/pdoom1-website/actions/runs/32001215197), branch `feedback/intake-contract`, 2026-08-17 06:20Z. `shivammathur/setup-php@v2` installed **PHP 8.2.33** and the suite executed against `public/ingest.php` — as far as the log shows, the first time that endpoint has run anywhere.

**All seventeen forced-failure cases hold.**

| | | | |
|---|---|---|---|
| F1 store dir unwritable | PASS | F12 payload over cap | PASS |
| F2 disk full mid-append | PASS | F13 rsync --delete over docroot | PASS |
| F3 fatal after write, before response | PASS | F14 store_root inside the docroot | PASS |
| F4 concurrent POSTs | PASS | F15 UTF-8 free text under cp1252 | PASS |
| F5 mail() returns false | PASS | F7 POST never leaves the client | PASS |
| F6 MTA accepts then discards | PASS | F8 response lost after the server wrote | PASS |
| F9 throttle trip | PASS | F16 localStorage quota exhausted | PASS |
| F10 honeypot filled | PASS | F17 hostile payload in every field | PASS |
| F11 malformed JSON | PASS | | |

Three summary lines from the same run:

- `OK: every claimed property was forced into its failing state and held.`
- `OK: every §9 hazard was forced and the refusal observed.`
- `OK: every mail-auth check was forced into its failing state and observed.`

Plus `M5 -- THE INTERLOCK, forced red (the reason this file exists)` inside the mail-auth job, which then passed — the interlock was exercised, not skipped.

## Why green rather than the expected red

The body predicted red on known defects. Those were being repaired in the `fix/intake-lane-*` branches, which are ancestors of this one, so the run measured the repaired state. **The green is earned rather than vacuous**: each case forces the failure and observes the refusal, which is this repo's own standard — a guard seen only in its passing state has not been shown to work.

The 92-line header on `test-ingest-destructive.py` is no longer a docstring. It is evidence, with a run id.

## Closing

The PR said *"Do not merge it at all yet"* and *"the result is the point"*. The result now exists and is recorded here, so the draft has served its purpose and is being closed rather than left open as ambient in-flight work. The branch and its 80 commits are untouched — reopen this, or cut a real PR for the parts that should land, whenever the remaining lanes are ready.

